PR #3, opened 2021-05-12
Fixed relative links in collections

applejag:
Some links were not being processed correctly by benbalter/jekyll-relative-links

Collections were disabled by default.
Now they work

https://github.com/benbalter/jekyll-relative-links#processing-collections


fredx30 (approved):
Seems legit.

Pikabanga (approved):

PR #4, opened 2021-05-12
RFC: core-lib-repo

applejag:
Proposed location to place utility code, such as loading config values or
serving version endpoints, so that our other repositories can take use of it.

Rendered: https://github.com/iver-wharf/rfcs/blob/90fdcad/_published/0004-core-lib-repo.md

### DIFF
--- a/_published/0004-core-lib-repo.md
+++ b/_published/0004-core-lib-repo.md
@@ -184,9 +184,11 @@ solution for [configuring ASP.NET Core apps](https://docs.microsoft.com/en-us/as
 
 1. What should be the name of the repo?
 
-   - `github.com/iver-wharf/wharf-core` *&larr; Suggested by this RFC*
-   - `github.com/iver-wharf/wharf-core-lib`
-   - `github.com/iver-wharf/wharf-extra`
-   - `github.com/iver-wharf/wharf-utility`
-   - `github.com/iver-wharf/wharf-util`
-   - `github.com/iver-wharf/wharf-utils`
+   | Suggestion                           | Votes                |
+   | ----------                           | -----                |
+   | github.com/iver-wharf/wharf-core     | @jilleJr, @Pikabanga |
+   | github.com/iver-wharf/wharf-core-lib |                      |
+   | github.com/iver-wharf/wharf-extra    |                      |
+   | github.com/iver-wharf/wharf-utility  |                      |
+   | github.com/iver-wharf/wharf-util     |                      |
+   | github.com/iver-wharf/wharf-utils    | @iverestefans        |
--- a/_published/0004-core-lib-repo.md
+++ b/_published/0004-core-lib-repo.md
@@ -1,0 +1,192 @@
+---
+layout: default
+# This is just to fool remark-stringify not to escape & symbols
+# See https://github.com/syntax-tree/mdast-util-to-markdown/issues/8
+shields_io_query_params: label=issue%20state&logo=github&style=flat-square
+
+# Update the following (it's YAML syntax)
+pr_id: 4 # Update this with PR number/ID. No leading zeros
+rfc_id: 0004 # Update this with PR number/ID. Use leading zeros
+rfc_feature_name: core-lib-repo # Use kebab-case
+title: "RFC-0004: core-lib-repo"
+rfc_author_username: jilleJr
+rfc_author_name: Kalle Jillheden # Or same as username, if you wish
+
+# Leave these. Collaborator changes this before merging
+impl_issue_id: 0
+impl_issue_repo: iver-wharf/wharf-api
+last_modified_date: YYYY-MM-DD
+---
+
+# {{page.title}}
+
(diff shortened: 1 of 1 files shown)


fredx30:
My gut feeling when downloading the repos was that this could use a master repo that defines the others as subrepos. This could potentially allow for a recursive clone of the master repo as well as providing an outer path where shared resources could be stored. Another such resource that comes to mind is the docker-compose repo/file.
Edit: See ref here https://git-scm.com/book/it/v2/Git-Tools-Submodules

_published/0004-core-lib-repo.md
```diff
@@ -0,0 +1,198 @@
...
+appVersion: v1.0.0
+buildGitCommit: 5971d3b585a722536730c39a22aa3148993f2985
+buildRef: 123
+buildDate: 2021-05-12T13:55:00+02:00
+```
+
+## Compatibility
+
+In architectural terms, this utility repository has to be "stable". Meaning it
+will be a hassle to update as there will be so many components relying on it.
```
Pikabanga, line 160: I mean, we shouldn't break things for the sake of it, but you can specify a certain version of the package when importing it right? So if you make a breaking change, it's not like it will all of a sudden break everything depending on it.

Pikabanga (approved):
+1 for creating a wharf-core repo

_published/0004-core-lib-repo.md
```diff
@@ -0,0 +1,198 @@
...
+1. What should be the name of the repo?
+
+   - `github.com/iver-wharf/wharf-core` *&larr; Suggested by this RFC*
+   - `github.com/iver-wharf/wharf-core-lib`
+   - `github.com/iver-wharf/wharf-extra`
+   - `github.com/iver-wharf/wharf-utility`
+   - `github.com/iver-wharf/wharf-util`
+   - `github.com/iver-wharf/wharf-utils`
+
+2. May parsing of the `.wharf-ci.yml` files live in this repo? Or should it get
```
iverestefans: Code connected with parsing is getting so complicated that it can have own repo IMO.

_published/0004-core-lib-repo.md
```diff
@@ -0,0 +1,198 @@
...
+
+It also allows for a unified way of configuring the services. If we can set a
+unified convention of setting configs via files and being able to override them
+with environment variables, then that would lift the operations-experience of
+using Wharf to a much better level. Heavily inspired by the `appsettings.json`
+solution for [configuring ASP.NET Core apps](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0)
+
+## Unresolved questions
+
+1. What should be the name of the repo?
```
iverestefans, line 185: I am voting for option `wharf-utils`

iverestefans (approved):

_published/0004-core-lib-repo.md
applejag: Indeed. What I tried to lift here was the potential issue of for example finding a bug or making a bigger change that has to then be reflected in 4-5 other repos in 4-5 PRs.

But with the assumption of versioning the core lib (which is my intention either way) then this issue get decreased to "we'll change it in due time anyway, no need to rush"

applejag:
> @fredx30
> My gut feeling when downloading the repos was that this could use a master repo that defines the others as subrepos. This could potentially allow for a recursive clone of the master repo as well as providing an outer path where shared resources could be stored. Another such resource that comes to mind is the docker-compose repo/file.
> Edit: See ref here https://git-scm.com/book/it/v2/Git-Tools-Submodules

Submodules is a fine thought. What you're proposing is a switch in how we work with the repos locally, where each local repo is a submodule of the "main" repo, if I understand you correctly.

Pros:

- Only 1 repo to clone, instead of 6 distinct. Simpler onboarding procedure, although that is just a one-time thing each developer has to do and [we do have it documented by now](https://iver-wharf.github.io/#/development/getting-started).
- 1 fewer repos, given that we put the core lib in this "main" repo
- No need to symlink the docker-compose.yml file to outside of the repo it's living in

Cons:

- We have to make sure the main repo is up to date all of the time with its submodule references
- Prone to confusion as we're versioning all the repos separately while the main repo that includes all the other repos has its own version of only the utility code inside itself.

Given these pros and cons, I personally would rather see the wharf-docker-compose repo have submodules and have some automatic script/GitHub action that updates its submodules whenever there's an update, instead of mixing this with the core/utility lib repo.

_published/0004-core-lib-repo.md
applejag: Hmm yea the subpackage cmd/pkg/core/wharfyml is up to 13 files as of now, and maybe we want to version it separately from the utility/core lib packages.

I'm removing it from this RFC

_published/0004-core-lib-repo.md
applejag: Resolving

_published/0004-core-lib-repo.md
applejag: I've added your vote to the RFC itself. You're currently outnumbered, but there's chance for more voters :)

applejag:
I'm finalizing the name as `wharf-core`

PR #6, opened 2021-05-12
RFC PR template

applejag:
Adds pull request template that prepopulates the description and labels with what we want


applejag:
I added lots of people to this PR. But I won't make it a habit.

fredx30 (approved):

Pikabanga (approved):

PR #11, opened 2021-05-26
RFC: Quay.io

applejag:
Hosting our built Docker images of our main API, web frontend, and provider APIs over at <https://quay.io/> by Red Hat.

## Rendered

https://github.com/iver-wharf/rfcs/blob/2e9fe0c/_published/0011-quay.io.md

### DIFF
--- a/_published/0011-quay.io.md
+++ b/_published/0011-quay.io.md
@@ -12,9 +12,9 @@ rfc_author_username: jilleJr
 rfc_author_name: Kalle Jillheden # Or same as username, if you wish
 
 # Leave these. Collaborator changes this before merging
-impl_issue_id: 0
-impl_issue_repo: iver-wharf/wharf-api
-last_modified_date: YYYY-MM-DD
+impl_issue_id: 49
+impl_issue_repo: iver-wharf/iver-wharf.github.io
+last_modified_date: 2021-05-27
 ---
 
 # {{page.title}}
--- a/_published/0011-quay.io.md
+++ b/_published/0011-quay.io.md
@@ -1,0 +1,121 @@
+---
+layout: default
+# This is just to fool remark-stringify not to escape & symbols
+# See https://github.com/syntax-tree/mdast-util-to-markdown/issues/8
+shields_io_query_params: label=issue%20state&logo=github&style=flat-square
+
+# Update the following (it's YAML syntax)
+pr_id: 11 # Update this with PR number/ID. No leading zeros
+rfc_id: 0011 # Update this with PR number/ID. Use leading zeros
+rfc_feature_name: quay.io # Use kebab-case
+title: "RFC-0011: quay.io"
+rfc_author_username: jilleJr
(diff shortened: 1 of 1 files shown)


Pikabanga (approved):
I don't see any issues here!

One thing to consider (which you did touch on), is that docker hub is way more popular, so we may lose some discoverability by not being there.

Would it be a crazy idea to host our images on both quay and docker hub?

applejag:
> @Pikabanga 
> One thing to consider (which you did touch on), is that docker hub is way more popular, so we may lose some discoverability by not being there.
> 
> Would it be a crazy idea to host our images on both quay and docker hub?

Hmm yea good point. But the added complexity of hosting it on multiple places for the sake of the few that actually search on Docker Hub instead is perhaps not a gain I'm personally willing to take. I'd rather we only have 1 dummy image pushed there that just redirects in its README.md to `quay.io/iver-wharf/...`, preferrably to how it's handled by .NET (https://hub.docker.com/_/microsoft-dotnet), node.js (https://hub.docker.com/_/node), PostgreSQL (https://hub.docker.com/_/postgres), et.al. in contrast to the /r/ paths for regular users, like Crystal (https://hub.docker.com/r/crystallang/crystal). No clue how difficult it is to get one of those "underscore" locations, probably by contacting Docker support or something

Pikabanga:
> > @Pikabanga
> > One thing to consider (which you did touch on), is that docker hub is way more popular, so we may lose some discoverability by not being there.
> > Would it be a crazy idea to host our images on both quay and docker hub?
> 
> Hmm yea good point. But the added complexity of hosting it on multiple places for the sake of the few that actually search on Docker Hub instead is perhaps not a gain I'm personally willing to take. I'd rather we only have 1 dummy image pushed there that just redirects in its README.md to `quay.io/iver-wharf/...`, preferrably to how it's handled by .NET (https://hub.docker.com/_/microsoft-dotnet), node.js (https://hub.docker.com/_/node), PostgreSQL (https://hub.docker.com/_/postgres), et.al. in contrast to the /r/ paths for regular users, like Crystal (https://hub.docker.com/r/crystallang/crystal). No clue how difficult it is to get one of those "underscore" locations, probably by contacting Docker support or something

I'm convinced to just go ahead with quay. The argument about discoverability is pretty weak anyway. People find things on github primarily :)

applejag:
> @Pikabanga 
> I'm convinced to just go ahead with quay. The argument about discoverability is pretty weak anyway. People find things on github primarily :)

Yea let's file that under *"we can add it later if we feel the need in the future"*

fredx30 (approved):
How do we publish images to quay? Does wharf itself have a module for that?

applejag:
> How do we publish images to quay? Does wharf itself have a module for that?

Wharf supports pushing to docker registries. We're using it internally to push to Harbor as they conform to the docker registry API specifications. Quay also conforms to those specs. All we need to do is just point Wharf to a different registry

PR #13, opened 2021-06-24
Authentication and Authorisation

fredx30:
### Description from the summary

Wharf should have authentication. Wharf currently has no authentication. This RFC will outline a plan to implement OpenID Connect (OIDC) in Wharf. The plan shall be limited to authentication and authorization with OIDC. User and group management will be offloaded onto the Identity provider this requires using Identity providers which embrace the concept of Roll-Based Access Control. The initial implementation will also be limited to only sourcing from 1 identity provider at a time, it will however be possible to extend this to multiple different sources later.

### Auth map V2
auth-map-v2


_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+  {{page.rfc_author_name}} ([@{{page.rfc_author_username}}](https://github.com/{{page.rfc_author_username}}))
+- Implementation
+  issue: [{{page.impl_issue_repo}}#{{page.impl_issue_id}}](https://github.com/{{page.impl_issue_repo}}/issues/{{page.impl_issue_id}})
+- Implementation
+  status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
+
+## Summary
+
+Wharf should have authentication. Wharf currently has no authentication. This
+will outline a plan to implement [OpenID Connect (OICD)](https://openid.net/connect/)
```
applejag: Isn't the abbreviation for [O]pen[ID] [C]onnect? (https://infosec.mozilla.org/guidelines/iam/openid_connect.html#common-abbreviations--definitions)

```suggestion
will outline a plan to implement [OpenID Connect (OIDC)](https://openid.net/connect/)
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+Some claim a good feature of secure infrastructure is access control. Currently,
+our only solution is network isolation. Being able to restrict access to varying
+degrees is a product goal to satisfy business needs.
+
+## Explanation
+
+Pre implementation stages
+
+1. OICD in any form demands full SSL security in all connections. In order to
```
applejag: Isn't the abbreviation for [O]pen[ID] [C]onnect? (https://infosec.mozilla.org/guidelines/iam/openid_connect.html#common-abbreviations--definitions)

```suggestion
1. OIDC in any form demands full SSL security in all connections. In order to
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+   avoid having to fill the development environment with --force-unsecure
+   bindings we should consider implementing certificate generation and importing
+   for dev environments.
+
+Implementations stages
+
+1. Docker compose setup for Hydra Proxy setup
+   - This involves integration with the currently existing DB.
+2. Hydra (Authorisation server) - Connect to IDP (Config for Azure AD iver.se)
```
applejag: We want to refrain from making references to any of our services. It does help with understanding for us, but as we [Iver devs] are the only ones that use the Iver AD then I suggest against referencing it.

Also, an explanation of the IDP abbreviation on its first occurrence would be helpful

Also 2, we're using american spelling :)

```suggestion
2. Hydra (Authorization server) - Connect to IDP (IDentity Provider), such as Azure AD
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+- Implementation
+  issue: [{{page.impl_issue_repo}}#{{page.impl_issue_id}}](https://github.com/{{page.impl_issue_repo}}/issues/{{page.impl_issue_id}})
+- Implementation
+  status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
+
+## Summary
+
+Wharf should have authentication. Wharf currently has no authentication. This
+will outline a plan to implement [OpenID Connect (OICD)](https://openid.net/connect/)
+in wharf. This RFC will only outline authentication with OICD and not
```
applejag: nit:

```suggestion
in Wharf. This RFC will only outline authentication with OIDC and not
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+pr_id: 12 # Update this with PR number/ID. No leading zeros rfc_feature_name:
+authentication # Use kebab-case title: "RFC-0012: Authentication" # Update this
+with PR number/ID and feature name. Use leading zeros rfc_author_username:
+fredx30 rfc_author_name: Fredrik Dyrvold # Or same as username, if you wish
+
+# Leave these. Collaborator changes this before merging
+
+impl_issue_id: 0 impl_issue_repo: iver-wharf/wharf-api last_modified_date:
+2021-06-16
```
applejag: Something seems to have happened to your frontmatter, and you're referencing PR 12 instead of PR 13

If your text editor made these changes then maybe it doesn't support frontmatters. If this is the case, then try look for some setting or plugin, or use a different editor when editing these RFCs

```suggestion
layout: default
# This is just to fool remark-stringify not to escape & symbols
# See https://github.com/syntax-tree/mdast-util-to-markdown/issues/8
shields_io_query_params: label=issue%20state&logo=github&style=flat-square

# Update the following (it's YAML syntax)
pr_id: 13 # Update this with PR number/ID. No leading zeros
rfc_feature_name: authentication # Use kebab-case
title: "RFC-0013: authentication" # Update this with PR number/ID and feature name. Use leading zeros
rfc_author_username: fredx30
rfc_author_name: Fredrik Dyrvold # Or same as username, if you wish

# Leave these. Collaborator changes this before merging
impl_issue_id: 0
impl_issue_repo: iver-wharf/wharf-api
last_modified_date: 2021-06-16
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+- RFC
+  PR: [iver-wharf/rfcs#{{page.pr_id}}](https://github.com/iver-wharf/rfcs/pulls/{{page.pr_id}})
+- Feature name: `{{page.rfc_feature_name}}`
+- Author:
+  {{page.rfc_author_name}} ([@{{page.rfc_author_username}}](https://github.com/{{page.rfc_author_username}}))
+- Implementation
+  issue: [{{page.impl_issue_repo}}#{{page.impl_issue_id}}](https://github.com/{{page.impl_issue_repo}}/issues/{{page.impl_issue_id}})
+- Implementation
+  status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
```
applejag: nit: Excess line wrapping

```suggestion
- RFC PR: [iver-wharf/rfcs#{{page.pr_id}}](https://github.com/iver-wharf/rfcs/pulls/{{page.pr_id}})
- Feature name: `{{page.rfc_feature_name}}`
- Author: {{page.rfc_author_name}} ([@{{page.rfc_author_username}}](https://github.com/{{page.rfc_author_username}}))
- Implementation issue: [{{page.impl_issue_repo}}#{{page.impl_issue_id}}](https://github.com/{{page.impl_issue_repo}}/issues/{{page.impl_issue_id}})
- Implementation status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+OpenID Connect - Flows
+A good guide on flows is available [here on auth0](https://auth0.com/docs/flows).
+Or for a more fun version checkout [this game by auth0](https://iverab-my.sharepoint.com/:i:/g/personal/fredrik_dyrvold_iver_se/Ef41bR71qeNOk-QWhCar0V4BZN8WqvM7iEf5g7ml9_s7Ew?e=ZoZRN6)
+
+### Identity provider
+
+We should connect to azure ad on our local deployments.
+
+We should not hav our own identity provider setup with wharf for now.
```
applejag: I think this is a bad idea. I'm a hard advocate for forcing the local dev environment to only rely on local resources.

We should look at some basic IDP that we can host alongside the docker-compose setup, so when you have the images downloaded then Wharf will not require any internet connection.

To me it feels silly to require all devs to setup something like a GitHub app just to log into the local dev Wharf instance, and it will get tedious or in our way fast.

I think that having one from the get-go should be a priority

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+   configured through `agular-auth-oicd-client`
+
+8. Wharf-provider-* - Implement and test
+   - Likely identical to procedure for wharf-api.
+
+### Flows
+
+OpenID Connect - Flows
+A good guide on flows is available [here on auth0](https://auth0.com/docs/flows).
+Or for a more fun version checkout [this game by auth0](https://iverab-my.sharepoint.com/:i:/g/personal/fredrik_dyrvold_iver_se/Ef41bR71qeNOk-QWhCar0V4BZN8WqvM7iEf5g7ml9_s7Ew?e=ZoZRN6)
```
applejag: Suggest embedding the image inside the repository instead of linking it to your sharepoint.

In #14, @Alexamakans have used a root folder named `assets`. Suggest doing the same.

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Identity provider
+
+We should connect to azure ad on our local deployments.
+
+We should not hav our own identity provider setup with wharf for now.
+
+We should also be able to publish code in such a way that the IDP is
+interchangeable. If it is used by the public they can set up their own IDP
+and connect to that.
```
applejag: What is the "it" that you're referring to?

Also, just throwing it out there, we may want to design this so we could support multiple IDP's for a single instance?

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Authorisation server
+
+There are many [alternatives](https://openid.net/developers/certified/) on
+alternative authorisation servers.
+
+We had a quick info share from spark the spark project. They are using open
+identity server 3 or 4. This however appears to be relatively tightly coupled
+with the .NET framework.
```
applejag: Don't refer to internal services.

```suggestion
We had a quick info share from coworkers of a different project. They are using open
identity server 3 or 4. This however appears to be relatively tightly coupled
with the .NET framework.
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+```
+ORY Hydra is a hardened, OpenID Certified OAuth 2.0 Server and OpenID Connect
+Provider optimized for low-latency, high throughput, and low resource
+consumption. ORY Hydra is not an identity provider (user sign up, user
+login, password reset flow), but connects to your existing identity provider
+through a login and consent app. Implementing the login and consent app in a
+different language is easy, and exemplary consent apps (Node) and SDKs for all
+common languages are provided.
+```
```
applejag: nit: Suggest using blockquote instead of codeblock.

I personally like to add the source link at the end of the citation instead of before it, as I find it easier to see where it's taken from.

```suggestion
project is "Ory - Hydra" ([docs](https://www.ory.sh/hydra/docs/)).
The descriptions of them from their GitHub at the present time is:

> ORY Hydra is a hardened, OpenID Certified OAuth 2.0 Server and OpenID Connect
> Provider optimized for low-latency, high throughput, and low resource
> consumption. ORY Hydra is not an identity provider (user sign up, user
> login, password reset flow), but connects to your existing identity provider
> through a login and consent app. Implementing the login and consent app in a
> different language is easy, and exemplary consent apps (Node) and SDKs for
> all common languages are provided.
>
> <https://github.com/ory/hydra>
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+They appear to be funded through opencollective and patreon.
```
applejag: This seems irrelevant. The source of funds should not be factored into our decision on which tool to use.

Also, companies usually tries to sell their products and support besides accepting donations/sponsoring.

```suggestion
```

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+Another alternative to consider could be [Keycloak](https://www.keycloak.org/).
+This appears to contain
+much more finished work such as a user admin panel for
+setting user and group rights. It is styled using themes. Con: Its a redhat
+project which means association with IBM which has had a certain downside in
+relation to opensource projects in the past see
```
applejag: This paragraph seems to be unfinished.

And RedHat, hate them or love them, has still made a lot of great contributions to the free and open source ecosystem. Their association with IBM in my opinion is not of concern today. We can start moving away from their tools, such as Quay, once they really start going against free and open source morals.

Without a source, this comment of yours seems to be based on rumors, and I suggest removing it.

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Compatibility
+
+Expect no compatability between versions of wharf without authentication to
+versions of wharf with authentication. There is a potential threat in leaving
+the system backwards compatible in that a non-secured version of the api could
+run together in an otherwise secure cluster.
```
applejag: I think having the authentication opt-in to begin with would be a good step, and then in the next major version jump we make it mandatory. Just so we can test the system with authentication so all parts play well first, before making the jump to mandatory auth.

There's a lot of microservices that needs to be authenticated, and in different ways. Seems very foot-shooting prone if we naively enforce OIDC

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+- User group management interfaces
+- Authorization
+- Permissions
+
+## Unresolved questions
+
+1. Swagger - how will this integrate with OpenID Connect auth?
+   some info is available at [swagger docs](https://swagger.io/docs/specification/authentication/openid-connect-discovery/)
+2. Production like deployments - We need to setup helm charts?, IDP
+   configuration for hydra?, and maybe more. [Some info available in docs](https://www.ory.sh/hydra/docs/production)
```
applejag: For Postgres we've decided that users deploy this themselves. Maybe we could do the same with hydra, depending on how many ways there are of hosting it. If it's "just" a basic pod, no PVC and no special requirements, then we could probably just embed it into our wharf-helm Helm chart

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+3. In oder to keep the docker containers able to start stateless/self-contained
+   we need to find a way to ensure that registering client with the
+   authorization server either becomes part of a run-once startup script for a
+   new deployment or gets properly built into the build process. [See docs](https://www.ory.sh/hydra/docs/guides/oauth2-public-spa-mobile)
```
applejag: I may misunderstand the question here, but this should arguably just be a config that you as an admin configure when deploying Wharf.

This is such a critical setting that I don't think it would be wise to configure dynamically inside the database, but instead we do it via config.yaml and environment variables

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+4. We need some way of introducing a first admin account. This will be needed
+   on database clean starts. Perhaps an SQL inject?
```
applejag: Observing how other services solve this, there's two main alternatives:

1. Receive some kind of error from the API telling us it does not have the initial admin account set up. Then the web frontend, or any other client accessing the API, can create an admin account. This gives better UX where the installation process is just `helm install` and then configure via the web interface.

2. We have a config in the deployment of something like `initialAdminUsername` and `initialAdminPassword` that is created when the API boots up and there are 0 users. This is a little worse UX as most work is done during the configuring before releasing for the first time, but better UX for IasC operations.

As we have most our configs declared beforehand, I find option 2 to be best, and also probably easier to implement anyway.

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+   -
+3. Wharf-web - configure and setup [angular-auth-oicd-client](https://github.com/damienbod/angular-auth-oidc-client)
+   - Implement login endpoint
+   - Implement consent endpoint
+   - Implement logout endpoint
+
+4. Wharf-api - Implement and test
+   - [Go SDK for ory Hydra](https://www.ory.sh/hydra/docs/sdk/go)
+   - Inspiration for [gin middleware here](https://github.com/janekolszak/gin-hydra/blob/master/ginhydra.go)
+     note this is outdated.
```
applejag: I'm having trouble understanding what ORY Hydra actually solves for us.

I think it would help with a graph (graphical/image or ASCII/Markdown art) that shows where ORY Hydra is meant to live.

1. Does the IDPs, such as Azure AD, talk to ORY Hydra or talk to our API then then talks to ORY Hydra?
2. Does ORY Hydra need any persistence? Do we give it direct access to our DB, or do our API act as a DB proxy, or does it have its own storage?
3. When someone wants to use our API as a service, do they also need the ORY Hydra SDK?

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+7. Edit wharf-web services to use bearer tokens through angular middleware
+   configured through `agular-auth-oicd-client`
+
+8. Wharf-provider-* - Implement and test
+   - Likely identical to procedure for wharf-api.
```
applejag: The providers today only forwards the "Authorization" header. If it receives a "401 Unauthorized" from the wharf-api then it returns the same response from the provider itself.

I think it would be wise to keep the providers dumb in this way. They do not manage any authentication/authorization at all.

_published/0012-authentication.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Future possibilities
+
+After this we should consider
+- User management interfaces
+- User group management interfaces
+- Authorization
+- Permissions
+
+## Unresolved questions
```
applejag: I know we'll get the question further on about LDAP or just using a in-database account store. How does this solution handle this?

Some of the best services, from the user and admin perspective, that we use are for example Grafana and GitLab that allows us to set up LDAP authentication as well as creating custom local users.

How does this proposed ORY Hydra solution handle this? For example mixing multiple login providers, such as username+password (for the initial admin account), an OIDC login (such as Azure AD or via GitHub Apps), and a LDAP login, all side by side (if the operations team so wanted it). Maybe even login from multiple LDAP or OIDC connections?

I sense that this is difficult to change later on if we end up only supporting OIDC and nothing else.

applejag (changes requested):
Lots of comments

_published/0012-authentication.md
fredx30: Do you have any tips on formatting these front matters? Right now the fall under generic markdown formatting. I cant keep spacing but i should be able to keep line breaks correctly. The paragraphs get auto merged when they are not double spaced to comply with the line length rules.

applejag: Depends on how you're formatting them. If manually, then leaving them as-is should be enough. If using [remark-lint](https://github.com/remarkjs/remark-lint), either via their CLI or via some editor integration, then make sure you've run the `npm install` as the `remark-frontmatter` is one of our dependencies. If using [Prettier](https://github.com/prettier/prettier-vscode) then it should be able to handle frontmatters, so make sure you're using the latest version.

Where is the formatting coming from?

_published/0012-authentication.md
fredx30: > I think this is a bad idea. I'm a hard advocate for forcing the local dev environment to only rely on local resources.

The concept of running locally without external connectivity seems highly valuable. 

> We should look at some basic IDP that we can host alongside the docker-compose setup, so when you have the images downloaded then Wharf will not require any internet connection.

In the ory environment this would require the setup of a local idp. This local idp (when sourced from ory) is called [Ory-Kratos](https://www.ory.sh/kratos). If we do this then i think we should put off the requirement of using the AzureAD till a later RFC ex. after authorisation is in place.

If we were to create/switch on a bypass for the authentication system based on a developement flag we may be able to get a prototype off the ground faster with a smaller change footprint. This could allow for continuing work on prototyping integration with AzureAD while still retaining non-internet-connected development. 

The azureAD still seems like a requirement for this feature as that is where Iver is at.

> To me it feels silly to require all devs to setup something like a GitHub app just to log into the local dev Wharf instance, and it will get tedious or in our way fast.

If we use azure or github as an identity provider then there is no need to setup anything more. The remote endpoints are login.microsoft.com or github.com depending on the github/azureAD choice.

> I think that having one from the get-go should be a priority

I agree we should be able to develop and use the dev environments locally without connectivity to the internet. I disagree with the assumption that we should have a local idp. Some disadvantages of having a local idp include
##### Sprawl of user administration
In a production environment with a separate idp of our own we would get extra work in the form of having to conform to usder off&on boarding procedures. Using our remote idp this would not be an issue as we would be piggypbacking off exsisting such infrastructure, past and present.
#####  GDPR Storage of additional user specific info
We will likely need to store the name of a user even if we only use the authorisation server. 
If we start hosting an IDP of our own we will also need to store email or other contact information in order to deal with account recovery options.
##### Complexity
If we need a local idp setup for wharf then that should be a project of its own that would need to get completed before the authentication project can be started. While getting a prototype user store up and running with ory kratos may be relativly simple- scaling that to production does not seem trivial. 
IMO Adding a requirement of a yet non-exsisting, non-planned idp would act as a full hard block/stop for this project.

_published/0012-authentication.md
fredx30: > What is the "it" that you're referring to?'

Wharf

```suggestion
We should also be able to publish code in such a way that the IDP is
interchangeable. If Iver-Wharf is used by the public they can set up their own IDP
and connect to that.
```

> Also, just throwing it out there, we may want to design this so we could support multiple IDP's for a single instance?

My understanding is that this is supported. Each separate endpoint should be one configuration file, as well as a link (auth icon) from the login interface. See sample from quay.io's login screen here 
image

_published/0012-authentication.md
fredx30: I think its possible to read microsoft style AD FS ldap attributes through OIDC scopes. There is an [article here](https://cloud.google.com/anthos/clusters/docs/on-prem/1.7/how-to/oidc-adfs#mapping_ldap_attributes_to_claim_names) that outlines a plan for how its done. Testing this for our use case which seems quite different requires setting up infrastructure that has costs (in both time an money) associated to it. I do not intend to test this in practice as of now.
If we use OIDC to connect an AD FS IdP to our hydra server then we are likely using LDAP i assume the same goes for azureAD. If we connect hydra to kratos (both are ory products) then we would not be using LDAP. Kratos uses postgres by default but has some other options.
If you happen to mean SAML instead of LDAP then the answer is no we would not support SAML as far as this RFC goes. The implementation is different. Use of OIDC is a starting assumption for this RFC. This is stated in the summary.

Multiple different OIDC endpoints are possible.

> I sense that this is difficult to change later on if we end up only supporting OIDC and nothing else.

Yes. I belive adding support for other authentication styles would be difficult counter productive if we achive compatability with OIDC. Use of OIDC is a starting assumption for this RFC. This is stated in the summary.

_published/0012-authentication.md
fredx30: As they are separate clients that connect to the untrusted zone which is the wharf-web-client then as far as my understanding goes they would have to individually establish trust with the authorization server.

Alternativly if we start and finish the project to move the provider clients (wharf-provider-github, wharf-provider-azuredevops, wharf-provider-gitlab) into the wharf-api such that they are not exposed microservices then the model could be simplified. 

1) Either they could be microservices that are unauthenticated but ARE restricted to ONLY the trusted network. In this case they could be kept outside the authentication system and just remain unprotected.

2) They could be turned into interfaces which get implemented by the wharf-api endpoint. This way they wouldnt be networked and therefore authentication would not be applicable.

_published/0012-authentication.md
fredx30: Ory Hydra is an implementation of the OpenID Connect standard. The standards documents themselves reccomend using an approved solution to implement the standard. Ory Hydra is one such approved solution.

See below for a map-
image
In this case the providers are just interfaces to the wharf-api and not exposed microservices for simplicity.

1. The wharf's environment link out to IdPs is the ory-hydra pod it acts as a gateway for all things identity.
2. Yes, for the tests i have done thus far i have been using postgres as persistent storage, but other options are available.
3. No but they will need to conform to OpenId Connect.

applejag: Very nice graphic, and fine answers! Would like to have this explained in the RFC in a nice way as well, so it answers my questions there. (I've not checked yet if you've already added them)

_published/0012-authentication.md
fredx30: The source i meant to find for this is [this post](https://blog.centos.org/2020/12/future-is-centos-stream/) from centos right after RedHat was aquired by IBM. In this case support was cut for a new image from May 31, 2029 to December 31, 2021.

_published/0012-authentication.md
applejag: While I do like the 2nd alternative more, Go does not have good plugin support. So option 1 is the best for allowing users to add more provider plugins without recompiling the main API. Letting them live as microservices also makes it easier to host them in k8s and docker-compose as then it will be as simple as adding yet another pod/container to the mix and then you're done (in theory).

I'll start working on an RFC for the reworked providers integration.

_published/0012-authentication.md
fredx30: The source of funds should absolutely be considered. In a worst case scenario you might end up working with an supporting cirme, terrorism, or other illegal activities. A more likely scenario is perhaps the introduction of a dependency developed by a mallicious third party.
Picking dependencies that remain supported and popular can save alot of future development work that comes with migrating dependencies.

_published/0012-authentication.md
applejag: Good answer!

So for the sake of this RFC, we could ignore LDAP. But supporting username+password locally, which as I understand it is possible with ORY Hydra, is more essential even if this is the initial proposal

_published/0012-authentication.md
fredx30: Although this should probably not mention the company specifically by name. I will instead replace this with the uncited motivation `Picking dependencies that remain supported and popular can save alot of future development work that comes with migrating dependencies.` and reformulate the surounding paragraph.

_published/0012-authentication.md
applejag: Ah great POV. Like if they're crowdfunded, then there's at least a community that confirms that it's not bad? I buy that

_published/0012-authentication.md
applejag: Awesome :call_me_hand:

_published/0012-authentication.md
applejag: Seems like I don't know the full implications of the term IDP.

I meant that for local development only we have some local username+password store that perhaps the ORY Hydra talks to. Then for production we only rely on third-party IDP such as Azure AD.

By having a local dev-only solution then we could really cut some corners and use more or less anything that's OpenID compliant. Could be a simple page with a single hardcoded username+password, or even just a button that says "Yep, log me in to local dev Wharf" without even having to provide any credentials. Doesn't even have to be a "certified" solution, as long as it complies with the spec so that we can manually e2e test it.

My intention is not to include some self-hosted IDP when we deploy Wharf to k8s, as that bring up so many of the problems that you talked about. Better to have this auth opt-in until we get that Azure AD integration set up.

fredx30:
Massive update! The change of design is dropping the middleware (hydra) infavour of a direct client to IDP approach. This should make everything simpler. Hydra does not appear to be meant to be used in the fashion i was trying to use it before.

_published/0012-authentication.md
fredx30: This is rewritten in the new version but may require an extra consideration again as the circumstances have changed.

fredx30:
Resolved lots of comments that are no longer relevant.

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+- Near perfectly mirrored compliance with top level organisational
+  security levels
+
+## Explanation
+
+By using the OIDC standard an authentication & authorisation system will be
+built in a minimalist fashion that offloads most of the work involved in
+administration to other pre-existing platforms.
+
+![Connectivity visualised](../assets/auth-map-v2.PNG)
```
applejag, line 86: Why does the Remote IDP talk directly to Wharf's database?

nit: Shouldn't the Remote IDP be placed in the "Wide area network" area? Like on the same side as the "Cronjob & Automation", but not on the same side as the "User Agent"

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+The OIDC protocol defines a token called an
+[ID Token](https://auth0.com/docs/tokens/id-tokens/id-token-structure).
+The ID token is signed by the source issuer. This signature is then used to
+validity and authenticity of the token. See
+[validating an OIDC ID-Token](https://connect2id.com/blog/how-to-validate-an-openid-connect-id-token).
+
+#### Sessions
+The wharf-api is currently stateless and does not use sessions. It is
+possible to either include the ID token in every request to the server or to
+start using sessions to store the ID token between requests.
```
applejag: This feels undecided. Would like an iron fist on this. Are we implementing sessions, are we using a third-party lib to deal with the sessions, or are we requesting a new ID token on each request?

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+pr_id: 13 # Update this with PR number/ID. No leading zeros
+rfc_feature_name: authentication & authorization # Use kebab-case
+title: "RFC-0013: authentication & authorization" # Update this with PR number/ID and feature name. Use leading zeros
+rfc_author_username: fredx30
+rfc_author_name: Fredrik Dyrvold # Or same as username, if you wish
+
+# Leave these. Collaborator changes this before merging
+
+impl_issue_id: 0 impl_issue_repo: iver-wharf/wharf-api last_modified_date:
+2021-06-16
```
applejag: nit:

```suggestion
impl_issue_id: 0
impl_issue_repo: iver-wharf/wharf-api
last_modified_date: YYYY-MM-DD
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### Sessions
+The wharf-api is currently stateless and does not use sessions. It is
+possible to either include the ID token in every request to the server or to
+start using sessions to store the ID token between requests.
+
+#### Local development vs production environments
+For the initial release of the feature the configuration will be tailored to
+using azure as an IDP. These settings will as necessary be exposed as
+environment variables to each respective docker image.
```
applejag, line 114: I think I understand what you mean with "to each respective docker image" here, but could you phrase it as explicit?

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+#### Choice of Client
+
+The OIDC standard defines a
+[list of certified clients](https://openid.net/developers/certified/)
+that can help simplify the implementation of ID Token retrieval.
+
+For the backend a full client will not be required. As the tokens just need
+to be validated and not requested. Validation can be done offline by caching
+the issuers' signature for future validations (OBS! The signature key may
+rotate on a timely basis).
```
applejag: Q: Is this where the session management is needed? Is not the ID Token a valid standalone auth source (except for the IDP's public cert used for validation)?

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+
+Expect no compatability between versions of wharf without authentication to
+versions of wharf with authentication. There is a potential threat in leaving
+the system backwards compatible in that a non-secured version of the api could
+run together in an otherwise secure cluster.
+
+We should aim to pick a solution where any IDP supporting OIDC and RBAC can be
+used. Locking ourselves to a single IDP when a standard such as OIDC exists
+seems like a  waste opportunity. Keeping the choice of IDP open will leave
+more room for choice and flexibility in the future.
```
applejag: Will this solution mean Wharf is incompatible with IDPs without RBAC support? Or will just the fine-grain user control be missed out? Like, can Wharf use an IDP without RBAC but then all users will be treated as sysadmins, or will that not work at all?

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Alternative solutions
+
+There are many types of identity verification systems.
+This RFC assumes that OIDC will be used. Within OIDC there are multiple
+different [flow definitions](https://developer.okta.com/docs/concepts/oauth-openid/#choosing-an-oauth-2-0-flow).
+We should primarily use `codeflow with pkce` as this is presumed to be the
+most secure, and best compatible at the time of writing.
```
applejag: Would be nice with a note here to explain "(best compatible) with what?", such as best compatible with our combination of SPA web and sole REST API access, or just because we have a SPA web?

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Future possibilities
+
+We may want to implement multiple different IDP sources. This could allow
+login from different sources than the Iver-Azure-AD. A sample situation
```
jilleJr, line 142: No internal references allowed, sorry

```suggestion
login from different sources than a single Azure-AD. A sample situation
```

This is not written down somewhere, except for this comment here and now, but we got directive to not make any company nor product references throughout the project, not even to Iver itself. Partially to make it seem more open and inviting, but more importantly as a hard line so we don't accidentally slip out references to confidential products or partners.

Iver relying on an Azure AD is not a peice of information that will necessarily collapse the company if it gets into the wrong hands, but it has been declared that we still should refrain from referencing the company where it's not necessary.

I can agree that it's kind of a flimsy rule, but I like it for the fact that it makes us less potent in accidentally exposing confidential or dangerous information.

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+where this could be useful is where/when we wish to give a client
+view/control over the deployment pipeline.
+
+When handling ID-Tokens from RBAC enabled IDPs its is possible to verify
```
applejag: nit: typo

```suggestion
When handling ID-Tokens from RBAC enabled IDPs it's possible to verify
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+custom permission levels using custom role claims.
+[See this link](https://docs.microsoft.com/en-us/azure/role-based-access-control/custom-roles)
+for more information about custom claims. This system allows for setup
+fine-tuned access control on a per-deployment basis. The initial setup
+for each of the different RBACs will likely require setup scripts on a
+per-provider basis as it requires interaction directly with each respective
+cloud on a non-standard basis.
```
applejag: nit: Would rather see this just documented instead of scripted. It should be such a one-time and only necessary when users needs more access. And just having some example docs of how to properly set this up with different IDPs so those with undocumented IDPs could connect the dots themselves.

Don't know exactly what you refer to with "setup scripts", so excuse me if I misunderstood

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+## Unresolved questions
+
+OIDC in any form demands full SSL security in all connections. In order to
+avoid having to fill the development environment with --force-unsecure
+bindings we should consider implementing certificate generation and importing
+for dev environments.
+
+Should we keep the wharf-api stateless and send the id-token in every
+request or should we create sessions and store the id token in each user's
+session?
```
applejag: Is the idea not to store the ID token on the client-side? Like in localStorage or in cookies?

Are ID-Tokens meant to only be seen by the server, or are they given to the server by the user? In the case of the latter, we could instead just rely on the user always providing the ID-token themselves on each request, letting them keep track of the session and keeping the wharf-api stateless.

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+for each of the different RBACs will likely require setup scripts on a
+per-provider basis as it requires interaction directly with each respective
+cloud on a non-standard basis.
+
+## Unresolved questions
+
+OIDC in any form demands full SSL security in all connections. In order to
+avoid having to fill the development environment with --force-unsecure
+bindings we should consider implementing certificate generation and importing
+for dev environments.
```
applejag: We can tell traefik to use a locally generated certificate. Ex: https://github.com/traefik/traefik/issues/5507#issuecomment-535603881. Need to be careful not to use the same cert across all us devs by storing it in the repo, as that would be an insane attack vector.

But instead we could have an extra step to the docker-compose setup, something like:

1. Generate cert using openssl, for example as defined in a `Makefile`
2. Trust the newly generated certificate, with different docs for windows and linux. For example in linux it would be something like this (which could also be inserted into a `Makefile`):
   ```sh
   sudo cp wharf-local-dev.crt /usr/local/share/ca-certificates/wharf-local-dev.crt
   sudo update-ca-certificates
   ```
3. ```sh
   docker-compose up
   ```
Voila

applejag (changes requested):
Just to be clear: would rather see my questions answered by altering the 00013-authentication.md, rather than only responding to my comments here on the PR. Just so the answers are documented properly :)

_published/0013-authentication.md
applejag: Actually, since it's a SPA web, we could just store it in memory and force a new ID-token on every page refresh. That way there's a smaller attack surface as no malicious code could find the token from cookies or localStorage.

Bonus point: if someone gets new roles in the IDP, all they have to do is refresh the page to have it apply

_published/0013-authentication.md
fredx30: Good tips.
The line connecting the IDP to the wharf DB was a remenant from the hydra setup. Its now removed.
Updated some other aspects aswell. It should be clearer now.
auth_map_v2

_published/0013-authentication.md
fredx30: Good point. I think we should keep the sessionless state of things. I belive sound design dictates statelessness whereever possible.

_published/0013-authentication.md
fredx30: Thanks, will try to fix the formatting issues related to this special meta info at a more final stage i have no way of highlighting these issues on a normal work basis.

_published/0013-authentication.md
fredx30: Explicit how? Listing the names of the names of the docker images does not seem relevant. Im still finding it relativly hard to deal with the particular loadup sequences on the different settings so i dont have an exact on how that will work. The reason i want to keep it relativly vauge is the lack of exacts untill the work is finished.

_published/0013-authentication.md
fredx30: The problem is the potential preformance risk of having an excessive number of requests fetching the IDP issuer keys.
No caching of the public keys for the IDP does not relate to sessions. Without caching the issuer public keys for the IDP a request to fetch would have to be made for every authorization request (or in the stateless session variety - every request). 
If we kept sessions then we would only have to authenticate once per session. This would almost entierly rid us of the edit: ~~imminent~~ minor preformance problem.

fredx30: Perhaps we shold a brainstomring meeting on this one.

_published/0013-authentication.md
fredx30: Requesting custom claims that dont exsist on a source IDP that cant define custom claims will always return false. So in our scenario we could get the id token but we could never get any form of higher priviledge.

fredx30: This could be a section to move into a finished discussion form in `alternative considered`.

_published/0013-authentication.md
fredx30: I believe the thought was current security practices. Implicit and hybrid code flows dont usually seem to be held in as high a regard for user agent authorisation. 
There is still some intresting testing to be done here for implicit trust towards implicitly trusted clients operating autonomously. There should probably be some info here on how id tokens and custom claims work. That also requires more testing.

_published/0013-authentication.md
fredx30: I understand the principals and values in this case the iver ad is public info if not explicitly written then atleast through dns definitions. In this case its the only motivator i know for picking azure ad the IDP. In all other aspects i believe auth0/okta is a better choice in wharf's case.

_published/0013-authentication.md
fredx30: At this point i imagine there will be just one custom role so the manual work of setting on role up wont be heavy. The scripts would have to serve as documentation if they were to be written. The scripts if written for instance for azure ad would for instance setup an app registration with a name and designate a certain user the app registration admin, then it could also initiate the app registration with a correct set of custom role(s - multiple at a later point), perhaps it would even be appropriate to ensure the correct securtiy settings regarding code flows are set.

_published/0013-authentication.md
fredx30: > Actually, since it's a SPA web, we could just store it in memory and force a new ID-token on every page refresh. That way there's a smaller attack surface as no malicious code could find the token from cookies or localStorage.
> Bonus point: if someone gets new roles in the IDP, all they have to do is refresh the page to have it apply

This is mostly outside of our control. This should be correctly handled by the client oidc framework, and should automagically do more or less what you want but through different mechanics. I would reffer to look at the mechanics of the refresh token in the oidc standard.

> Is the idea not to store the ID token on the client-side? Like in localStorage or in cookies?

Yes but the question here is should we for localStorage or rely on sessionStorage. Other choices seem to be much more difficult as these are the two that are obviously supported by the OIDC client framework. One difference is how susceptible the sessions are to cors attacks. It should result in a bit of a different user experience between the two in terms of how well a user can stay logged in between different types of shutdown. After my last discussion with a colleague im leaning towards retaning the default setting of sessionStorage for increased protection against CORS attacks.

> Are ID-Tokens meant to only be seen by the server, or are they given to the server by the user? In the case of the latter, we could instead just rely on the user always providing the ID-token themselves on each request, letting them keep track of the session and keeping the wharf-api stateless.

Typically ID tokens are not supposed to be seen by the server as the whole point of the original design spec of OIDC is to allow a familiar authentication/authorisation flow without having to fully disclose ones identity. However with the introduction of RBACs, and their use in custom claims its possible to setup ID tokens in such a way that they dont GDPR relevant userdata while still being fully functional as authorization level proof much like an access code would in the OAUTH spec.

_published/0013-authentication.md
```diff
@@ -0,0 +1,165 @@
...
+
+Some claim a good feature of secure infrastructure is access control. Currently,
+our only solution is network isolation. Being able to restrict access to varying
+degrees is also a product goal to satisfy business needs.
+
+Tying the user management of wharf to central repositories such as
+organisation wide IDPs is good for a number of reasons.
+- Centrally managed on/off boarding:
+   - Leads to less administration
+   - Roll based security inheritance allows automatic updates of user access
```
agmo: Ro**ll** based -> Ro**le**-based?

_published/0012-authentication.md
fredx30: This particular scenario has changed quite a bit since the rewrite. In the new model i propse we include a flipswitch to disable token validation in the backend for developer environments. This could allow navigation as normal entierly without logging in.

_published/0013-authentication.md
fredx30: The id token and access token are stored by the frontend for as long as allowed untill the refresh token is used to request a new set. For now that timeout is 1 hour. The access token is sent to the backend on every request. The id token is kept by the frontend for visuals such as seeing what user you logged in as.

_published/0013-authentication.md
fredx30: The limitations here dont seem to be technical as much as beurocratic. On the technical side this is meant to enable even the most complex and secure flows making it easier to convince the beurocrats the tech is ok to implement.
I ran the tests on creating the tokens autonomously as a trusted client and it appears to work in most fashions. Once the tokens are recieved the flow towards the api is identical to the web. The custom roles appear to work as well but production environment final tests are still blocked.

fredx30: The new rewording is slightly more abiguous to allow this to be solved in implementation/configuration rather than stated in the RFC.

_published/0013-authentication.md
fredx30: I have reworded this such that its left up to the implementation. 

The decision on which IDP to use for authentication will need to be made and customised in setup per instance.

Since this is not a one-time setup i would personally have this setup be automated rather than manual. Personally i also belive code can serve as documentation. The setup is likely needed to be done one per instance of wharf.

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+The OIDC protocol defines a token called an
+[Access Token](https://auth0.com/docs/security/tokens/access-tokens).
+The access token is signed by the source issuer. This signature is then used to
+validity and authenticity of the token. See
+[validating an OIDC Access Token](https://auth0.com/docs/security/tokens/access-tokens/validate-access-tokens).
+
+### Session Management
+
+The wharf-api is currently stateless and does not use sessions. It is
+possible to either include the token in every request to the server.
```
Alexamakans: missing second option following "either"

Either
```diff
- possible to either include the token in every request to the server.
+ possible to either include the token in every request to the server, or [option two].
```
or remove "either".

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+## Dictionary
+
+- OIDC - [OpenID Connect (OIDC)](https://openid.net/connect)
+
+- IDP - Identity provider - some examples include
+  [Google Workspace](https://workspace.google.com/intl/sv),
+  [Azure AD](https://azure.microsoft.com/en-us/services/active-directory), and
+  [Auth0/Okta](https://www.okta.com/)
+
+- RBAC - [Roll-Based Access Control](https://auth0.com/docs/authorization/rbac/)
```
Alexamakans: typo:
```suggestion
- RBAC - [Role-Based Access Control](https://auth0.com/docs/authorization/rbac/)
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+- Implementation status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
+
+## Summary
+
+Wharf should have authentication. Wharf currently has no authentication.
+This RFC will outline a plan to implement OpenID Connect (OIDC) in Wharf.
+The plan shall be limited to authentication and authorization with
+OIDC. User and group management will be offloaded onto the Identity provider
+this requires using Identity providers which embrace the concept of
+Roll-Based Access Control. The initial implementation will also be limited
```
Alexamakans: typo:
```suggestion
Role-Based Access Control. The initial implementation will also be limited
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+degrees is also a product goal to satisfy business needs.
+
+Tying the user management of wharf to central repositories such as
+organisation wide IDPs is good for a number of reasons.
+
+- Centrally managed on/off boarding:
+
+  - Leads to less administration
+
+  - Roll based security inheritance allows automatic updates of user access
```
Alexamakans: typo:
```suggestion
  - Role-based security inheritance allows automatic updates of user access
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+A OIDC type access token can be used for authentication. The custom claims
+(requires an IDP with support for RBAC) can be used to specify access levels.
+To keep the scope manageable the task of dividing up and refining access
+levels and permission will be left to a later date. For this RFC all kinds
+of read permissions won't require valid tokens, but all write type requests
+will require valid ID-tokens that have been signed by a pre-approved issuer.
+The OIDC protocol defines a token called an
+[Access Token](https://auth0.com/docs/security/tokens/access-tokens).
+The access token is signed by the source issuer. This signature is then used to
+validity and authenticity of the token. See
```
Alexamakans:
```suggestion
validate the authenticity of the token. See
```
or
```suggestion
check the validity and authenticity of the token. See
```
?

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+[list of certified clients](https://openid.net/developers/certified/)
+that can help simplify the implementation of ID Token retrieval.
+
+For the backend a full client will not be required. As the tokens just need
+to be validated and not requested. See
+[validating an access token](https://auth0.com/docs/security/tokens/access-tokens/validate-access-tokens)
+
+## Compatibility
+
+Expect no compatability between versions of wharf without authentication to
```
Alexamakans: typo:
```suggestion
Expect no compatibility between versions of wharf without authentication to
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+versions of wharf with authentication. There is a potential threat in leaving
+the system backwards compatible in that a non-secured version of the api could
+run together in an otherwise secure cluster.
+
+We should aim to pick a solution where any IDP supporting OIDC and RBAC can be
+used. Locking ourselves to a single IDP when a standard such as OIDC exists
+seems like a  waste opportunity. Keeping the choice of IDP open will leave
```
Alexamakans: typo:
```suggestion
seems like a wasted opportunity. Keeping the choice of IDP open will leave
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,171 @@
...
+This RFC assumes that OIDC will be used. Within OIDC there are multiple
+different [flow definitions](https://developer.okta.com/docs/concepts/oauth-openid/#choosing-an-oauth-2-0-flow).
+We should primarily use `codeflow with pkce` as this is presumed to be the
+most secure at the time of writing.
+
+## Future possibilities
+
+We may want to implement multiple different IDP sources. This could allow
+login from different IDPs. A sample situation where this could be useful is
+when a third party view/control over the deployment pipeline.
```
Alexamakans: Some words seem to be missing here in the second sentence, around "when a third party view/control over the".

Alexamakans (changes requested):
I think it looks solid, really nice work 👍

As for the unresolved question of stateless vs. stateful:
I would like to cast my vote for the wharf-api staying stateless. I don't know a whole lot about the security implications of either, though.

Found some typos and seemingly missing parts in sentences.

_published/0013-authentication.md
fredx30: Good plan! Will write this into a todo for the wharf-docker-compose repo as a subtask of auth.

fredx30: Perhaps the domain name `wharf.local` might be a good starting point aswell. Probably an easy switch later in implementation.

_published/0013-authentication.md
fredx30: I guess techinically the authenticity is gained from the `aud` attribute which has to be checked separately when the whole token has been had its signature validated. Going with  `validate the token.`

_published/0013-authentication.md
fredx30: Now - `A sample situation where this could be useful is
when an external third party should be granted view or admin rights to the
deployment pipeline.`

_published/0013-authentication.md
fredx30: Rewrote the section - `A sample situation where this could be useful is
when an external third party should be granted view or admin rights over the
deployment pipeline.`

_published/0013-authentication.md
fredx30: I have done som poking around the problem on this one and this seems to be a very minor problem which is probably best solved later and left as simple as possible now.

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+# Leave these. Collaborator changes this before merging
+
+impl_issue_id: 0
+impl_issue_repo: iver-wharf/wharf-api
+last_modified_date: YYYY-MM-DD
+---
+
+# {{page.title}}
+
+- RFC PR: [iver-wharf/rfcs#{{page.pr_id}}](https://github.com/iver-wharf/rfcs/pulls/{{page.pr_id}})
```
applejag: Fix in URL from https://github.com/iver-wharf/rfcs/pull/19

```suggestion
- RFC PR: [iver-wharf/rfcs#{{page.pr_id}}](https://github.com/iver-wharf/rfcs/pull/{{page.pr_id}})
```

_published/0013-authentication.md
applejag: Having to add `wharf.local` into your `hosts` file seems like extra unneeded work. Think we should register the local certs to simply `localhost`. Maybe a more unique port than `:5000`, but still.

Is there something particular you're thinking of for going with `wharf.local` instead of `localhost`?

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+
+- IDP - Identity provider - some examples include
+  [Google Workspace](https://workspace.google.com/intl/sv),
+  [Azure AD](https://azure.microsoft.com/en-us/services/active-directory), and
+  [Auth0/Okta](https://www.okta.com/)
+
+- RBAC - [Role-Based Access Control](https://auth0.com/docs/authorization/rbac/)
+
+- Issuer - In this RFC this refers to the issuer field in a OIDC metadata
+  sheet. The issuer is an authority for validating JWT tokens used in OIDC.
```
applejag: nit: "JWT tokens" = "JSON web token tokens"

```suggestion
  sheet. The issuer is an authority for validating JWTs used in OIDC.
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+Wharf should have authentication. Wharf currently has no authentication.
+This RFC will outline a plan to implement OpenID Connect (OIDC) in Wharf.
+The plan shall be limited to authentication and authorization with
+OIDC. User and group management will be offloaded onto the Identity provider
+this requires using Identity providers which embrace the concept of
+Role-Based Access Control. The initial implementation will also be limited
+to only sourcing from 1 identity provider at a time, it will however be
+possible to extend this to multiple different sources later.
+
+## Dictionary
```
applejag, line 44: Kudos for adding this "Dictionary" section!

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+![Connectivity visualised](../assets/auth-map-v2.PNG)
+
+### The Access Token
+
+A OIDC type access token can be used for authentication. The custom claims
+(requires an IDP with support for RBAC) can be used to specify access levels.
+To keep the scope manageable the task of dividing up and refining access
+levels and permission will be left to a later date. For this RFC all kinds
+of read permissions won't require valid tokens, but all write type requests
+will require valid ID-tokens that have been signed by a pre-approved issuer.
```
applejag, line 95: So all "read" API operations will be available anonymously? Did I understand that correctly?

If that's the case, then why not require valid tokens for all endpoints? In what way does this keep the scope more manageable?

_published/0013-authentication.md
applejag: Re-reading it I understand now better than when I wrote the comment, and no longer find it an issue in the formulation.

Marking as resolved.

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+For the initial release of the feature the configuration will be tailored to
+using azure as an IDP. These settings will as necessary be exposed as
+environment variables to each respective docker image.
+
+For use in offline developer environments it should be possible to switch on a
+developer mode switch that allows bypassing validation of access
+tokens such that a valid login is not required.
+
+
+### SSL Config Required as a part of OIDC
```
applejag: lint: two empty lines, should only be 1

```suggestion
tokens such that a valid login is not required.

### SSL Config Required as a part of OIDC
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+
+For OIDC ssl is required. SSL certs will be inserted into the traefik config
+and generated with openSSL as part of the repo setup.
+
+### Choice of Client
+
+The OIDC standard defines a
+[list of certified clients](https://openid.net/developers/certified/)
+that can help simplify the implementation of ID Token retrieval.
```
applejag, line 130: The plan is to use https://github.com/damienbod/angular-auth-oidc-client right? I think it's better to specify that one here.

```suggestion
The OIDC standard defines a
[list of certified clients](https://openid.net/developers/certified/)
that can help simplify the implementation of ID Token retrieval.

For the frontend, <https://github.com/damienbod/angular-auth-oidc-client>
will be used.
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+For the backend a full client will not be required. As the tokens just need
+to be validated and not requested. See
+[validating an access token](https://auth0.com/docs/security/tokens/access-tokens/validate-access-tokens)
+
+## Compatibility
+
+Expect no compatibility between versions of wharf without authentication to
+versions of wharf with authentication. There is a potential threat in leaving
+the system backwards compatible in that a non-secured version of the api could
+run together in an otherwise secure cluster.
```
applejag, line 144: Currently we actually support Basic Auth. While this OIDC solution is incompatible with Basic Auth, I do think it would be wise to let users choose between Basic Auth, OIDC, and potentially completely unauthorized.

Enforcing OIDC and removing the two other (less secure) existing options would mean some customers can not upgrade Wharf.

I like the idea of deprecating the unauthorized alternative, and perhaps even the Basic Auth one, but not from the get-go. A bunch of deprecation warnings until non-OIDC solutions are completely blocked is how I envision such a transition at least.

Mixing auth solutions for the same Wharf instance/installation should of course not be supported.

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+seems like a wasted opportunity. Keeping the choice of IDP open will leave
+more room for choice and flexibility in the future.
+
+## Alternative solutions
+
+There are many types of identity verification systems.
+This RFC assumes that OIDC will be used. Within OIDC there are multiple
+different [flow definitions](https://developer.okta.com/docs/concepts/oauth-openid/#choosing-an-oauth-2-0-flow).
+We should primarily use `codeflow with pkce` as this is presumed to be the
+most secure at the time of writing.
```
applejag: IMO this could be more decisive.

```suggestion
We will use `codeflow with pkce` as this is presumed to be the
most secure at the time of writing.
```

_published/0013-authentication.md
```diff
@@ -0,0 +1,177 @@
...
+## Future possibilities
+
+We may want to implement multiple different IDP sources. This could allow
+login from different IDPs. A sample situation where this could be useful is
+when an external third party should be granted view or admin rights to the
+deployment pipeline.
+
+When handling tokens from RBAC enabled IDPs it's possible to verify
+custom permission levels using custom role claims.
+[See this link](https://docs.microsoft.com/en-us/azure/role-based-access-control/custom-roles)
```
applejag, line 168: The page has a warning message:

> Adding a management group to AssignableScopes is currently in preview. This preview version is provided without a service level agreement, and it's not recommended for production workloads. /.../
>
> https://docs.microsoft.com/en-us/azure/role-based-access-control/custom-roles

Is this something we need to worry about? Have you tested this feature, or is it hypothetical?

(You're not required to have tested this feature, as it's for a future plan and not of major importence right now. I'm merely curious)

_published/0013-authentication.md
applejag: > making it easier to convince the beurocrats the tech is ok to implement.

Lol

> The new rewording is slightly more abiguous to allow this to be solved in implementation/configuration rather than stated in the RFC.

:+1: 

I'm glad with the explanation. Resolving.

_published/0013-authentication.md
applejag: Alright! Convinced :)

_published/0013-authentication.md
applejag: My point there about only storing it in memory was that if we store the tokens in "memory", i.e. local/scoped javascript variables, instead of localStorage/sessionStorage then that makes it protected against XSS attacks, and since it's an SPA then the client won't need to make that many requests to the remote IDP anyway but instead only when opening a new tab or refreshing the page.

And that way if your remote account from the IDP has gotten more access roles then to refresh the client's token they only need to refresh their page to get the new access roles, as that will force the client to obtain a new token.

> Typically ID tokens are not supposed to be seen by the server /.../ its possible to setup ID tokens in such a way that they dont GDPR relevant userdata while still being fully functional as authorization level proof much like an access code would in the OAUTH spec.

That sh*t's baller :call_me_hand:

fredx30: JS variables are not a simple alternative here there may be a custom way of implementing it but since it would require alot more work i would see the present as a choice between local and session storage. Per meeting we belive sessionstorage is better than localstorage because the user can in a simpler way wipe the cache (opening/closing tab).

_published/0013-authentication.md
fredx30: Potentially add a frontend warning if communicating on http non-secure

fredx30: Given localhost is possible that seems much better.

_published/0013-authentication.md
fredx30: Per discussion- A valid login will be required for all and any access. Custom roles need to be investigated further and should perhaps have their own RFC.

_published/0013-authentication.md
fredx30: I agree with this. I think there will/should be an on/off switch for OIDC. If OIDC is off then basic auth could be enabled or disabled separately. This could cause some strange issues if both are enabled but dosent seem to be a real relevant problem that needs asap handling.

_published/0013-authentication.md
fredx30: I have tested this in parts. Its hidden behind the azure premium paywall. Its executable through auth0/okta iaas.

_published/0013-authentication.md
fredx30: This is semi rewritten for the subtitle  `### SSL Config Required as a part of OIDC` now.

applejag (approved):
LGTM!

Alexamakans (approved):

PR #14, opened 2021-06-28
RFC: Rework handling of test results

Alexamakans:
Changing handling of test results in wharf-api.

Parses the .trx files once and insert into database instead of parsing each time it's requested.
Also adding the possibility to get more detailed summaries, as well as more detailed individual test results.

## Rendered

https://github.com/iver-wharf/rfcs/blob/master/_published/0014-test-results-handling.md


_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+Having the tests' error messages is pretty much a necessity when trying
+to fix them.
+
+## Related GitHub issues
+- wharf-api [#11](https://github.com/iver-wharf/wharf-api/issues/11) [#17](https://github.com/iver-wharf/wharf-api/issues/17)
+- wharf-web [#13](https://github.com/iver-wharf/wharf-web/issues/13)
+
+## Explanation
+
+<details><summary>wharf-api</summary>
```
applejag: Excessive use of `<details>` tag. Makes it harder to read. Suggest to just using a level-3 heading instead.

```suggestion
### wharf-api
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+
+The POST `/build/{buildid}/artifact` endpoint handles inserting artifacts.
+If there are TRX (XML) files, it also parses them to create an array of `TestResult` and
+one `TestResultSummary` per file.
+
+The summaries get inserted into the database table `test_result_summary`.
+The results get inserted into the database table `test_result`.
+
+![Database structure](https://github.com/iver-wharf/rfcs/blob/33e14dceb44755ce5e0c66b257bbffe2bb3568d7/_assets/wharf-db-graph.png)
```
applejag: Suggest referring to local file instead of the remote file

```suggestion
![Database structure](../_assets/wharf-db-graph.png)
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    StartedOn   *time.Time   `gorm:"nullable;default:NULL" json:"startedOn" format:"date-time"`
+    CompletedOn *time.Time   `gorm:"nullable;default:NULL" json:"finishedOn" format:"date-time"`
+    GitBranch   string       `gorm:"size:300;default:'';not null" json:"gitBranch"`
+    Environment null.String  `gorm:"nullable;size:40" json:"environment" swaggertype:"string"`
+    Stage       string       `gorm:"size:40;default:'';not null" json:"stage"`
+    Params      []BuildParam `gorm:"foreignKey:BuildID" json:"params"`
+    IsInvalid   bool         `gorm:"not null;default:false" json:"isInvalid"` 
+    // added 
+    TestResultSummaryCount uint `gorm:"not null" json:"testResultSummaryCount"`
+}
```
applejag: tip: Perhaps use `diff` language to show modifications instead? Like so?

```diff
@@ database_models.go @@
 type Build struct {
     BuildID     uint         `gorm:"primaryKey" json:"buildId"`
     StatusID    BuildStatus  `gorm:"not null" json:"statusId"`
     ProjectID   uint         `gorm:"not null;index:build_idx_project_id" json:"projectId"`
     Project     *Project     `gorm:"foreignKey:ProjectID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
     ScheduledOn *time.Time   `gorm:"nullable;default:NULL" json:"scheduledOn" format:"date-time"`
     StartedOn   *time.Time   `gorm:"nullable;default:NULL" json:"startedOn" format:"date-time"`
     CompletedOn *time.Time   `gorm:"nullable;default:NULL" json:"finishedOn" format:"date-time"`
     GitBranch   string       `gorm:"size:300;default:'';not null" json:"gitBranch"`
     Environment null.String  `gorm:"nullable;size:40" json:"environment" swaggertype:"string"`
     Stage       string       `gorm:"size:40;default:'';not null" json:"stage"`
     Params      []BuildParam `gorm:"foreignKey:BuildID" json:"params"`
     IsInvalid   bool         `gorm:"not null;default:false" json:"isInvalid"` 
+    TestResultSummaryCount uint `gorm:"not null" json:"testResultSummaryCount"`
 }

+type TestResultSummary struct {
+    ArtifactID  uint	  `gorm:"not null;index:testresultsummary_idx_artifact_id" json:"artifactId"`
+    Artifact    *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+    BuildID     uint      `gorm:"not null;index:testresultsummary_idx_build_id" json:"buildId"`
+    Build       *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+    RunCount    uint	  `gorm:"not null" json:"runCount"`
+    SkipCount   uint	  `gorm:"not null" json:"skipCount"`
+    FailCount   uint	  `gorm:"not null" json:"failCount"`
+    PassCount   uint	  `gorm:"not null" json:"passCount"`
+}

//etc..
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+</details>
+
+<details><summary>wharf-web</summary>
+
+wharf-web changes to use the new GET
+`/build/{buildid}/test-result-summaries` and GET `/build/{buildid}/artifact/{artifactid}/test-results`
+endpoints to retrieve the test result data instead of using the existing
+GET `/build/{buildid}/tests-results` endpoint.
```
applejag: nit: I sometimes find it more readable to include the HTTP method inside the inline code block

> wharf-web changes to use the new GET `/build/{buildid}/test-result-summaries` and GET `/build/{buildid}/artifact/{artifactid}/test-results` endpoints to retrieve the test result data instead of using the existing GET `/build/{buildid}/tests-results` endpoint.

vs

> wharf-web changes to use the new `GET /build/{buildid}/test-result-summaries` and `GET /build/{buildid}/artifact/{artifactid}/test-results` endpoints to retrieve the test result data instead of using the existing `GET /build/{buildid}/tests-results` endpoint.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+
+We also want to display more detailed test result summaries in wharf-web.
+For this we will need to also store the messages that the tests produce.
+At the very least for failed tests.
+
+## Motivation
+
+This is bad for the user because it leads to slower response times when
+loading the project view. It's also bad for the API because it leads to
+unnecessary fetches, and pointless repeated computation.
```
applejag: nit: I find this sentence ambiguous. Is the proposal from this RFC bad? Suggest making each paragraph more standalone in its context/subjects.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+- Implementation issue: [{{page.impl_issue_repo}}#{{page.impl_issue_id}}](https://github.com/{{page.impl_issue_repo}}/issues/{{page.impl_issue_id}})
+- Implementation status: ![GitHub issue state](https://img.shields.io/github/issues/detail/state/{{page.impl_issue_repo}}/{{page.impl_issue_id}}?{{page.shields_io_query_params}})
+
+## Summary
+
+The intent is to avoid filtering and parsing test result files (TRX)
+from the generated artifacts each time we want to display them or a
+summary. In wharf-web's project view, the Status field for
+builds (Failed/Completed) currently does this for each build every
+time it's refreshed. This leads to unnecessary requests and computation.
```
applejag: You're digging into what's wrong with the current implementation here. Focus on what this proposal is about instead, and leave the arguments for the "Motivation" section

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+
+This is bad for the user because it leads to slower response times when
+loading the project view. It's also bad for the API because it leads to
+unnecessary fetches, and pointless repeated computation.
+
+Having the tests' error messages is pretty much a necessity when trying
+to fix them.
+
+## Related GitHub issues
+- wharf-api [#11](https://github.com/iver-wharf/wharf-api/issues/11) [#17](https://github.com/iver-wharf/wharf-api/issues/17)
```
applejag: nit: Missing empty line after heading

```suggestion
## Related GitHub issues

- wharf-api [#11](https://github.com/iver-wharf/wharf-api/issues/11) [#17](https://github.com/iver-wharf/wharf-api/issues/17)
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+- wharf-web [#13](https://github.com/iver-wharf/wharf-web/issues/13)
+
+## Explanation
+
+<details><summary>wharf-api</summary>
+
+The POST `/build/{buildid}/artifact` endpoint handles inserting artifacts.
+If there are TRX (XML) files, it also parses them to create an array of `TestResult` and
+one `TestResultSummary` per file.
```
applejag: Don't know where we talked about this before internally, but it should be a separate endpoint to upload test results files.

Such as `POST /build/{buildid}/tests-results` that operates the same as the `POST /build/{buildid}/artifact` but it also parses the upload.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    ScheduledOn *time.Time   `gorm:"nullable;default:NULL" json:"scheduledOn" format:"date-time"`
+    StartedOn   *time.Time   `gorm:"nullable;default:NULL" json:"startedOn" format:"date-time"`
+    CompletedOn *time.Time   `gorm:"nullable;default:NULL" json:"finishedOn" format:"date-time"`
+    GitBranch   string       `gorm:"size:300;default:'';not null" json:"gitBranch"`
+    Environment null.String  `gorm:"nullable;size:40" json:"environment" swaggertype:"string"`
+    Stage       string       `gorm:"size:40;default:'';not null" json:"stage"`
+    Params      []BuildParam `gorm:"foreignKey:BuildID" json:"params"`
+    IsInvalid   bool         `gorm:"not null;default:false" json:"isInvalid"` 
+    // added 
+    TestResultSummaryCount uint `gorm:"not null" json:"testResultSummaryCount"`
```
applejag: This field seems frivolous. Suggest to instead have a ["Has Many" association](https://gorm.io/docs/has_many.html) here

```suggestion
    TestResultSummaries []TestResultSummary `gorm:"foreignKey:BuildID" json:"testResultSummaries"`
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    if len(testResults.Results) > 0 {
+    	testResults.Count = len(testResults.Results)
+    	testResults.ArtifactID = artifactId
+    } else {
+    	// dbnotfound error
+    	return
+    }
+    
+    // 200 with testResults
+}
```
applejag: You say that this is psuedo code, but it's still filled with implementation details IMO (the whole code section, not just the part I'm highlighting).

It makes it a little harder to review this RFC.

We've commonly said that the "## Explanation" section should be written as if it was documentation on an already existing feature. Maybe that will help?

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+endpoints to retrieve the test result data instead of using the existing
+GET `/build/{buildid}/tests-results` endpoint.
+
+There would also be a way to view a build's test result details. [#17](https://github.com/iver-wharf/wharf-api/issues/17)
+</details>
+
+## Compatibility
+
+This breaks backward compatibility with projects using the
+GET `build/{buildid}/tests-results` endpoint, since it will get removed.
```
applejag: There's no prior mention of it being removed. This seems like an easy enough backward compatability change.

Every time we break backward compatability we have to jump a major version and stuff breaks.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Unresolved questions
+
+- I am having trouble thinking of how to test filling out the database
+  with data from the old test results. **What would be a good way to do this?**
+  
+- **Is an upload endpoint, separate from the one for other artifacts, for test
+  results necessary?**
+  I can see it being necessary if somebody wants to upload local test results
+  or something, but it doesn't feel like that would be required. ref. to: [Create a separate method to upload test results apart from artifacts](https://github.com/iver-wharf/wharf-api/issues/11)
```
applejag, line 257: What we aimed with two different endpoints is that we don't want to overload an endpoint for just uploading raw artifacts with a lot of parsing logic.

And then we just happen to store the test results as artifacts as well.

Using the same endpoint would force us to add a lot of query parameters as feature flags, such as
- when a user wants to upload a .trx file but don't want the test results to be included in the build's score;
- or when user uploads a binary that happens to have .trx file extension, should that artifact upload be rejected because of failed to parse or just silently ignore it?;
- or when we want to support other test results formats, there ought to be some test results that happens to have the same file extension

Mostly just to separate the responsabilities of the endpoints

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+
+- As mentioned in [#17](https://github.com/iver-wharf/wharf-api/issues/17), we
+  are unlikely to have to store successful test details.
+  **Is there any foreseeable drawback to going that route?**
```
applejag: Don't think so. We could add a feature-flag for it either on the endpoint as a query parameter or as a global config on the Wharf API level.

More nobs with sensible defaults isn't that bad

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+Nothing comes to mind.
+
+## Future possibilities
+
+Nothing comes to mind.
+
+## Unresolved questions
+
+- I am having trouble thinking of how to test filling out the database
+  with data from the old test results. **What would be a good way to do this?**
```
applejag, line 251: Ignore it. Old test results uploaded previously should be ignored. Otherwise they collide with the other notes I wrote about below regarding "when a user wants to upload a .trx file but don't want the test results to be included in the build's score", as we cannot know for sure and I'd declare there's no real need for it to parse old test results.

Did I understand you correctly?

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+}
+// new
+func parseTRX(file *File) []TestResult, TestResultSummary {
+    // ...
+}
+```
+
+```go
+// build.go
+// new, /build/{buildid}/test-result-summaries
```
applejag: IMO it should instead be named `GET /build/{buildid}/test-results-summary` and have it return them all summed up.

We could still have per-artifact summary as well. Maybe something like this?:

```json
GET /build/42/test-results-summary
{
  "failed": 3,
  "passed": 5,
  "skipped": 0,
  "artifacts": [
    {
      "artifactId": 123,
      "filename": "MyResults1.trx",
      "failed": 1,
      "passed": 3,
      "skipped": 0
    },
    {
      "artifactId": 124,
      "filename": "MyResults2.trx",
      "failed": 3,
      "passed": 2,
      "skipped": 0
    }
  ]
}
```

Don't really see the need for having a "count" property returned. At least in our environments of both Go and JS/TS we have that value pre-calculated for us inside the arrays themselves.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    	Results     []TestResult `json:"results"`
+    	ArtifactID  uint `json:"artifactId"`
+    	Count       uint `json:"count"`
+    }
+    
+    testResults := TestResults{}
+	
+    m.Database.
+        Where(&TestResult{BuildID: buildId, ArtifactID: artifactId}).
+        Find(&testSummaries.Summaries)
```
applejag:
```suggestion
        Find(&testResults.Results)
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    
+    if len(testResults.Results) > 0 {
+    	testResults.Count = len(testResults.Results)
+    	testResults.ArtifactID = artifactId
+    } else {
+    	// dbnotfound error
+    	return
```
applejag: Better to return an empty list when there are no results, as the artifact may still exist but it happens to not have any test results.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    
+    testResults := TestResults{}
+	
+    m.Database.
+        Where(&TestResult{BuildID: buildId, ArtifactID: artifactId}).
+        Find(&testSummaries.Summaries)
+    
+    if len(testResults.Results) > 0 {
+    	testResults.Count = len(testResults.Results)
+    	testResults.ArtifactID = artifactId
```
applejag: The caller should have the artifact ID handy as it's a parameter to the endpoint. No need in returning it here

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    FailCount   uint	  `gorm:"not null" json:"failCount"`
+    PassCount   uint	  `gorm:"not null" json:"passCount"`
+}
+// new
+type TestResult struct {
+    ArtifactID  uint 	  `gorm:"not null;index:testresult_idx_artifact_id" json:"artifactId"`
+    Artifact    *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+    Name        string	  `gorm:"not null;" json:"name"`
+    Ran         string 	  `gorm:"not null;" json:"ran"`
+    Passed      string 	  `gorm:"not null;" json:"passed"`
```
applejag: How does these properties, "Ran" & "Passed", work? They're both strings. What do they represent?

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    // added 
+    TestResultSummaryCount uint `gorm:"not null" json:"testResultSummaryCount"`
+}
+// new
+type TestResultSummary struct {
+    ArtifactID  uint	  `gorm:"not null;index:testresultsummary_idx_artifact_id" json:"artifactId"`
+    Artifact    *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+    BuildID     uint      `gorm:"not null;index:testresultsummary_idx_build_id" json:"buildId"`
+    Build       *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
+    RunCount    uint	  `gorm:"not null" json:"runCount"`
```
applejag: nit: "Total" would be more suiting IMO. "Run" is somewhat unconventional word

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,256 @@
...
+    SkipCount   uint	  `gorm:"not null" json:"skipCount"`
+    FailCount   uint	  `gorm:"not null" json:"failCount"`
+    PassCount   uint	  `gorm:"not null" json:"passCount"`
```
applejag: nit: Personal preference towards "skipped", "failed", & "passed"; instead of "skipCount", "failCount", "passCount".

Maybe "-Count" is clearer?

applejag (changes requested):

_published/0014-test-results-handling.md
Alexamakans: I imagined RunCount to be the sum of FailCount and PassCount, leaving out SkipCount.
Then TotalCount would be RunCount + SkipCount.

Would it be better to instead have RunCount be all 3s total, and then rename it TotalCount?

_published/0014-test-results-handling.md
Alexamakans: Typo, meant to be bools.

Alexamakans:
```suggestion
    Ran         bool 	  `gorm:"not null;" json:"ran"`
    Passed      bool 	  `gorm:"not null;" json:"passed"`
```

_published/0014-test-results-handling.md
Alexamakans: Since they're asking for the test-results, doesn't it make sense to say that they weren't found?
Or is it that it doesn't count as an error because the request was successful and it's a valid state?

applejag: In the API we've only returned 404 so far when otherwise only returning a single object. As soon as a list of objects is returned then an empty array has been returned instead of 404.

Mostly because of implementation details, but I'd say better to stay conventional and then we can fix the other endpoints behaviour later.

_published/0014-test-results-handling.md
applejag: Suggest then to instead have an enum for it, something like "state" or "status" with the enum values of "skipped", "passed", & "failed"

_published/0014-test-results-handling.md
Alexamakans: I think Skipped/Failed/Passed could be interpreted as bools, especially with Passed being a bool in TestResult.
I don't like the Count suffix a lot either, but when it comes to writing SQL I think it's better to be more verbose than not.

Could make sense to at least change them to past tense, keeping the "-Count".

_published/0014-test-results-handling.md
applejag: Yea. For me at least I understand the "total" nomenclature much better than "run". Also makes more sense that if a value is a sum then it's the sum of all others, not just some others.

We're talking ints here, so like 4 bytes per test result artifact. We can afford it, and even more fields if we so want to.

_published/0014-test-results-handling.md
Alexamakans: I see.

Should I rewrite it to be more like documentation then?

_published/0014-test-results-handling.md
applejag: (We're not writing SQL though)

_published/0014-test-results-handling.md
applejag: I personally find it helpful when trying to explain at the right level of abstraction, so I vote for it. Striking the right balance is always hard

_published/0014-test-results-handling.md
Alexamakans: That does seem better, yeah.

Picked the Count thing up when I was working in the azure repo, but that was when receiving data 😜

_published/0014-test-results-handling.md
Alexamakans: Yup! 👍🏻

_published/0014-test-results-handling.md
applejag: It's not a terrible idea. But it's inconsistent with how most other endpoints work in Wharf.

We don't have a REST styleguide, so it's hard to know :)

_published/0014-test-results-handling.md
Alexamakans: Does this mean that the actual TRX file is posted to both endpoints?

Or that it gets posted to the artifact endpoint first, then a request could be made to process it through the
"upload test result" endpoint, using the artifact id?

_published/0014-test-results-handling.md
Alexamakans: Writing was a poor choice of words, but it does end up in an SQL database.
Either way, I think it helps in avoiding confusion.

_published/0014-test-results-handling.md
Alexamakans: Roger.

_published/0014-test-results-handling.md
Alexamakans: I've decided to change it to [ Total, Failed, Passed, Skipped ], the more I thought about it the more annoying the "-Count" got.

_published/0014-test-results-handling.md
applejag: Only the "upload test result" endpoint is used. If you as a user wants to upload other artifacts, you use the "upload artifact" endpoint. Then it's an implementation detail that we store the uploaded files to both of the endpoints as identical artifacts, while only the "upload test result" files are interpreted, and we might store them differently in the future such as its own tab of "Test result artifacts" vs "regular artifacts"

Pikabanga: I agree that there should be a separate endpoint for uploading test results, that way the upload artifact endpoint can remain "dumb".

Pikabanga:
I think we should try to model this to support other test result formats too. TRX is just one of many formats.

Not that I don't think this database model won't work for other formats too, but it could be worth looking into some other common test result formats to see if nothing obvious has been missed.

One thing to consider is to add a mandatory parameter to the upload test results endpoint, to specify format. That way we don't need to do any file extension checking or content parsing to determine the format.

For example, TRX has support for attached files for a specific test. Is that something we want to support (and display in the UI).

Here you can see the test result formats that the Azure Pipeline task Upload Test Results supports: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=trx%2Cyaml

Edit: Just to be clear, I think it's perfectly fine to only support .trx in version 1, but I think the design should be made with multiple formats supported in mind.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,257 @@
...
+  with data from the old test results. **What would be a good way to do this?**
+  
+- **Is an upload endpoint, separate from the one for other artifacts, for test
+  results necessary?**
+  I can see it being necessary if somebody wants to upload local test results
+  or something, but it doesn't feel like that would be required. ref. to: [Create a separate method to upload test results apart from artifacts](https://github.com/iver-wharf/wharf-api/issues/11)
+
+- As mentioned in [#17](https://github.com/iver-wharf/wharf-api/issues/17), we
+  are unlikely to have to store successful test details.
+  **Is there any foreseeable drawback to going that route?**
```
Pikabanga, line 264: Yes, you may want to see successful tests to track execution time over multiple runs (performance regression). I also think it should store which tests have been skipped so you can easily see if some test has been skipped dynamically by mistake.

Pikabanga (changes requested):

_published/0014-test-results-handling.md
applejag: May want to make this opt-in. There's a lot of strings there (test names) we may not be interested in. The failed and skipped tests are a minority among the total number of tests, so they can stay in by default IMO.

It would be nice to be able to query the history of a certain test though, that's a very tempting and useful feature.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+func parseTRX(file *File) []TestResult, TestResultSummary {
+  // ...
+}
+```
+
+The new endpoint `GET /build/{buildid}/test-results-summary` fetches
+all test result summaries for a build and constructs a json response
+that contains them, and has a summary of them, like:
+
+<details><summary>GET /build/42/test-results-summary</summary>
```
applejag: No damage in having a lengthy RFC. Suggest using subheaders instead of `<details>`. If it's worth hiding then it's not worth including at all, IMO. (edge cases may apply)

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+
+Nothing comes to mind.
+
+## Future possibilities
+
+Nothing comes to mind.
+
+## Unresolved questions
+
+Nothing comes to mind.
```
applejag, line 246: (Moving @Pikabanga's discussion over here)

> I think we should try to model this to support other test result formats too. TRX is just one of many formats.
> 
> Not that I don't think this database model won't work for other formats too, but it could be worth looking into some other common test result formats to see if nothing obvious has been missed.

I personally find this current proposal quite extensible. We could consider attachments or other test fail states (such as the [MSTest's "inconclusive" state](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/ms244275(v=vs.120)) in contrast to "skipped") later if we want to. I'm OK with leaving that for whenever we want to start supporting something like Java's JUnit, JS Jest, Go's built in test runner, et.al.

> One thing to consider is to add a mandatory parameter to the upload test results endpoint, to specify format. That way we don't need to do any file extension checking or content parsing to determine the format.
>
> /.../
>
> Edit: Just to be clear, I think it's perfectly fine to only support .trx in version 1, but I think the design should be made with multiple formats supported in mind.

I disagree with having a mandatory parameter, but an optional one would be benefitial.

Such as
- `POST /build/{buildid}/test-results`: Guess based on file extension, or error if unknown
- `POST /build/{buildid}/test-results?format=TRX`: Try parse as TRX, no matter what the file extension is.

Implementation wise it would just be a hardcoded map.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+
+- wharf-api [#11](https://github.com/iver-wharf/wharf-api/issues/11) [#17](https://github.com/iver-wharf/wharf-api/issues/17)
+- wharf-web [#13](https://github.com/iver-wharf/wharf-web/issues/13)
+
+## Explanation
+
+### wharf-api
+
+The `POST /build/{buildid}/artifact` endpoint handles inserting artifacts, same
+as before. However, if it's a TRX file, one should call the
```
applejag: nit: Should be clear that only supporting TRX is not the goal here.

```suggestion
as before. However, if it's a test results file (eg: .trx), one should call the
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+}
+// new, /build/{buildid}/artifact/{artifactid}/test-results
+func (m ArtifactModule) getBuildArtifactTestResultsHandler(c *gin.Context) {
+  // fetch test results for the specified test
+}
+// new
+func parseMultipartFormData(c *gin.Context) []*File {
+  // ...
+}
+// new
```
applejag: nit; All functions here are now, so I think it would be better to write that above the code snippet in free text instead of in each comment.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+- `GET /build/{buildid}/artifact/{artifactid}/test-results`\
+  To get test result details for build.
+  
+- `POST /build/{buildid}/test-results`\
+  To upload test result files for build.
+
+Deprecated endpoint
+- `GET /build/{buildid}/tests-results`
+
+There would also be a way to view a build's test results' details. [#17](https://github.com/iver-wharf/wharf-api/issues/17)
```
applejag, line 213: How do you propose this is done? I suggest its own tab in the build details page, but maybe you have a different view? 

Want details here. Perhaps a rough scetch or markdown-formatted sample of how it should be structured. Ex:

> ## <kbd>Details</kbd> | <kbd>Logs</kbd> | `Tests` :red_circle: | <kbd>Artifacts</kbd>
>
> - :green_circle: Passed: 256
> - :yellow_circle: Skipped: 0
> - :red_circle: **Failed: 3**
>
> ### Failed tests
>
> > #### Some/Sample/Tests
> >
> > ```
> > Error: Strings did not match
> >   Expected: Foo
> >     Actual: Bar
> > ```

Just to show what kind of info should be presented to the user

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+      "skipped": 0
+    }
+  ]
+}
+```
+</details>
+
+### wharf-web
+
+wharf-web changes to use the new endpoints
```
applejag, line 193: Think a note on how test results should be displayed in the list of builds should be handled would be good here. For example noting that the test results in a new "test results" column should be summed up from the `testResultSummaries` property on each build retrieved from the `GET /projects/{projectid}/builds` endpoint

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+  To get summary of all test result summaries for build.
+  
+- `GET /build/{buildid}/artifact/{artifactid}/test-results`\
+  To get test result details for build.
+  
+- `POST /build/{buildid}/test-results`\
+  To upload test result files for build.
+
+Deprecated endpoint
+- `GET /build/{buildid}/tests-results`
```
applejag: How should the deprecation of this endpoint be handled? Both regarding the backend AND frontends

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+### wharf-api
+
+The `POST /build/{buildid}/artifact` endpoint handles inserting artifacts, same
+as before. However, if it's a TRX file, one should call the
+`POST /build/{buildid}/test-results` endpoint instead. This tells the server
+to parse them as well, creating several `TestResult`s and one `TestResultSummary`
+per TRX file.
+
+The summaries get inserted into the database table `test_result_summary`.\
+The results get inserted into the database table `test_result`.
```
applejag: In the issues you've linked to we've used the terminology "Test summary vs test details". Suggest using the same here in both the RFC nomenclature as well as the table name, i.e. `test_result` &rarr; `test_result_detail`

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+```
+```go
+// artifact.go
+// new
+type File struct {
+  name string
+  fileName string
+  data []bytes
+}
+// new, /build/{buildid}/test-results
```
applejag: Suggest to instead use the swaggo syntax here, as this is Go code anyway

```suggestion
// @router /build/{buildid}/test-results [put]
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+func (m ArtifactModule) putTestResultsHandler(c *gin.Context) {
+  // parseMultipartFormData to get the files
+  // for each file, parse and store in db
+}
+// new, /build/{buildid}/artifact/{artifactid}/test-results
```
applejag: Suggest to instead use the swaggo syntax here, as this is Go code anyway

```suggestion
// @router /build/{buildid}/artifact/{artifactid}/test-results [get]
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+</details>
+
+### wharf-web
+
+wharf-web changes to use the new endpoints
+- `GET /build/{buildid}/test-results-summary`\
+  To get summary of all test result summaries for build.
+  
+- `GET /build/{buildid}/artifact/{artifactid}/test-results`\
+  To get test result details for build.
```
applejag: To not lock ourselves too much into that a test result is an artifact, I suggest instead something like `GET /build/{buildid}/test-results/{testresultid}` _(gosh I dislike that only-lowercase format, but that's an RFC for another day)_

Would also be benefitial with an endpoint with aggregated results from all endpoints on a certain build, as that's most probably how the API will be used anyway as we'll query the test result details for an entire build and not have to query for artifacts first and then have the "n+1" issue again that we have with builds.

Such as: `GET /build/{buildid}/test-result-details`

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,220 @@
...
+  
+- `POST /build/{buildid}/test-results`\
+  To upload test result files for build.
```
applejag: Maybe for unity with the naming difference, we should name this endpoint as `test-result-details` as well?

applejag (changes requested):
Now I had the chance to look more in-depth in the design. Left some comments

_published/0014-test-results-handling.md
Pikabanga: > Implementation wise it would just be a hardcoded map.

Sure, I was more thinking of scenarios where there are multiple formats that store the results as .xml or .json. Then you would have to parse the content to determine the format. That was the complexity I wanted to avoid with the format parameter.

But sure, we can do a best effort to determine type (like TRX based on file extension), and throw an error if we are unable to.

applejag: Oh yea, 100% agree. Attempting to guess based on content just stinks unexpected behaviour and overall bad UX. Filename is a little easier, for those cases we actually can conclude them.

_published/0014-test-results-handling.md
Alexamakans: How about
1. `GET /build/{buildid}/test-result-details`
    - Fetches all test result details for all tests for build.
2. `GET /build/{buildid}/test-result-details/{testresultid}` (i.e. testresultsummaryid, artifactid?)
    - Fetches specific test result details for specific test for build.
3. `GET /build/{buildid}/test-results-summary`
    - Fetches summary of summaries, along with the summaries, for build.
4. `GET /build/{buildid}/test-results-summary/{testresultsummaryid}` (i.e. artifactid)
    - Fetches specific summary for build.

Does that seem right?

`{testresultid}` confused me a little because
- Test results, henceforth test result details, don't have their own unique id currently.
- Test result summaries don't have their own name for their unique id, but since it's a 1-to-1 with Artifact we can use the artifact id.

_(Yeah, and it doesn't get better when it's as long as it is here either 😂)_

_published/0014-test-results-handling.md
Alexamakans: Not entirely sure about this, since both details and summary get parsed (or generated, if the format uploaded doesn't include a summary like the current format does). Perhaps `POST /build/{buildid}/test-result-data`, as that is what's actually getting sent to the API.

applejag: Yea, miss by me. In my head I pictured that the data you're uploading and downloading looks the same. But TRX isn't even close to our JSON format. Mb!

_published/0014-test-results-handling.md
applejag: Sure. Think the 4th endpoint there is a bit frivolous though. You already get that data when querying the 3rd endpoint. And would much rather see the `POST` endpoint return the summary so users don't need to query the one they just uploaded

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,231 @@
...
++  Skipped     uint      `gorm:"not null" json:"skipped"`
++}
+
++type TestResultDetail struct {
++  ArtifactID  uint              `gorm:"not null;index:testresult_idx_artifact_id" json:"artifactId"`
++  Artifact    *Artifact         `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
++  Name        string	        `gorm:"not null" json:"name"`
++  Message     string            `gorm:"nullable" json:"message"`
++  StartedOn   *time.Time        `gorm:"nullable;default:NULL;" json:"startedOn" format:"date-time"`
++  CompletedOn *time.Time        `gorm:"nullable;default:NULL;" json:"finishedOn" format:"date-time"`
```
applejag: nit: we ought to be using the `null` package for nullable fields instead of pointers.

Also, inconsistent json property name.

```suggestion
+  StartedOn   null.Time         `gorm:"nullable;default:NULL;" json:"startedOn" format:"date-time"`
+  CompletedOn null.Time         `gorm:"nullable;default:NULL;" json:"completedOn" format:"date-time"`
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,231 @@
...
++  Failed      uint      `gorm:"not null" json:"failed"`
++  Passed      uint      `gorm:"not null" json:"passed"`
++  Skipped     uint      `gorm:"not null" json:"skipped"`
++}
+
++type TestResultDetail struct {
++  ArtifactID  uint              `gorm:"not null;index:testresult_idx_artifact_id" json:"artifactId"`
++  Artifact    *Artifact         `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
++  Name        string	        `gorm:"not null" json:"name"`
++  Message     string            `gorm:"nullable" json:"message"`
```
applejag: nit:

```suggestion
+  Message     null.String       `gorm:"nullable" json:"message"`
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,231 @@
...
+func (m ArtifactModule) getBuildAllTestResultDetailsHandler(c *gin.Context) {
+  // fetch all test result details for build
+}
+// @router /build/{buildid}/test-result-details/{testresultsummaryid} [get]
+func (m ArtifactModule) getBuildTestResultDetailsHandler(c *gin.Context) {
+  // fetch all test result details for specific test for build
+}
+// @router /build/{buildid}/test-results-summary [get]
+func (m ArtifactModule) getBuildTestResultsSummary(c *gin.Context) {
+  // fetch summary of summaries, along with the summaries, for build
```
applejag: I find this phrase hard to interpret. Maybe something like this?

```suggestion
  // fetch sum of each artifact's test result summary for build
  // as well as the list of each artifact's test result summary
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,231 @@
...
++  Build       *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
++  Total       uint      `gorm:"not null" json:"total"`
++  Failed      uint      `gorm:"not null" json:"failed"`
++  Passed      uint      `gorm:"not null" json:"passed"`
++  Skipped     uint      `gorm:"not null" json:"skipped"`
++}
+
++type TestResultDetail struct {
++  ArtifactID  uint              `gorm:"not null;index:testresult_idx_artifact_id" json:"artifactId"`
++  Artifact    *Artifact         `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
```
applejag: Would personally prefer a CASCADE here instead of RESTRICT

```suggestion
+  Artifact    *Artifact         `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE" json:"-"`
```

Docs: https://gorm.io/docs/constraints.html#Foreign-Key-Constraint & https://www.postgresql.org/docs/9.5/ddl-constraints.html#DDL-CONSTRAINTS-FK

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,231 @@
...
+  Params      []BuildParam `gorm:"foreignKey:BuildID" json:"params"`
+  IsInvalid   bool         `gorm:"not null;default:false" json:"isInvalid"` 
++ TestResultSummaries []TestResultSummary `gorm:"foreignKey:BuildID" json:"testResultSummaries"`
+}
+
++type TestResultSummary struct {
++  ArtifactID  uint      `gorm:"not null;index:testresultsummary_idx_artifact_id" json:"artifactId"`
++  Artifact    *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
++  BuildID     uint      `gorm:"not null;index:testresultsummary_idx_build_id" json:"buildId"`
++  Build       *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
```
applejag: Would personally prefer a CASCADE here instead of RESTRICT

```suggestion
+  Artifact    *Artifact `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE" json:"-"`
+  BuildID     uint      `gorm:"not null;index:testresultsummary_idx_build_id" json:"buildId"`
+  Build       *Build    `gorm:"foreignKey:BuildID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE" json:"-"`
```

Docs: https://gorm.io/docs/constraints.html#Foreign-Key-Constraint & https://www.postgresql.org/docs/9.5/ddl-constraints.html#DDL-CONSTRAINTS-FK

applejag (changes requested):

_published/0014-test-results-handling.md
applejag: Maybe deserves its own RFC regarding the `null.Time` as we've used `*time.Time` elsewhere. You're free to resolve with only the json change:

```suggestion
+  StartedOn   *time.Time        `gorm:"nullable;default:NULL;" json:"startedOn" format:"date-time"`
+  CompletedOn *time.Time        `gorm:"nullable;default:NULL;" json:"completedOn" format:"date-time"`
```

_published/0014-test-results-handling.md
applejag: This is a change I'd rather see in the entire repo. I'll make an RFC on it for where it's RESTRICT elsewhere, but I think it would be more suiting to have CASCADE here to begin with.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,255 @@
...
++  Passed      uint      `gorm:"not null" json:"passed"`
++  Skipped     uint      `gorm:"not null" json:"skipped"`
++}
+
++type TestResultDetail struct {
++  TestResultDetailID uint       `gorm:"primaryKey" json:"testResultDetailId"`
++  ArtifactID  uint              `gorm:"not null;index:testresult_idx_artifact_id" json:"artifactId"`
++  Artifact    *Artifact         `gorm:"foreignKey:ArtifactID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
++  BuildID     uint              `gorm:"not null;index:testresult_idx_build_id" json:"buildId"`
++  Build       *Build            `gorm:"foreignKey:BuildID;constraint:OnUpdate:RESTRICT,OnDelete:RESTRICT" json:"-"`
```
Alexamakans:
```suggestion
+  Build       *Build            `gorm:"foreignKey:BuildID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE" json:"-"`
```

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,274 @@
...
+- `GET /build/{buildid}/tests-results`
+
+There would also be a way to view a build's test results' details. [#17](https://github.com/iver-wharf/wharf-api/issues/17)
+Something like this,
+
+> ## Details | Logs | `Tests` :red_circle: | Artifacts
+>
+> - :green_circle: Passed: 256
+> - :yellow_circle: Skipped: 0
+> - :red_circle: **Failed: 3**
```
applejag: nit: (purely cosmetic changes)

Need to [add `jemoji`](https://docs.github.com/en/enterprise/2.13/user/articles/emoji-on-github-pages) then it works *at least a little bit*:

image

Seems like the https://www.webfx.com/tools/emoji-cheat-sheet/ is the source of truth here, and the green and yellow circle emojies are not yet added.

Suggest changing to something like:

```suggestion
> ## Details | Logs | `Tests` :x: | Artifacts
>
> - :heavy_check_mark: Passed: 256
> - :large_orange_diamond: Skipped: 0
> - :x: **Failed: 3**
```

image

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,274 @@
...
+The `POST /build/{buildid}/artifact` endpoint handles inserting artifacts, same
+as before. However, if it's a test results file (eg: .trx), one should call the
+`POST /build/{buildid}/test-results` endpoint instead. This tells the server
+to parse them as well, creating several `TestResultDetail`s and one `TestResultSummary`
+per TRX file.
+
+The summaries get inserted into the database table `test_result_summary`.\
+The results get inserted into the database table `test_result_detail`.
+
+![Database structure](../\_assets/wharf-db-graph.png)
```
applejag: Because `npm run lint-fix` added this backslash in the path then the URL becomes invalid as Jekyll does not unescape URLs:

image

(for reference, `%5C` is the escape sequence for backslash `\`)

The underscore folders are special folders to Jekyll. The ones in this repo are considered "collections", as defined in the `_config.yml` and [has to start with an underscore](https://jekyllrb.com/docs/collections).

Suggest renaming this new `_assets` directory to simply `assets` and then updating this path like so:

```suggestion
![Database structure](../assets/wharf-db-graph.png)
```

applejag (changes requested):
Jekyll-specific changes :)

_published/0014-test-results-handling.md
Alexamakans: Ah, I see.

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Motivation
+
+The Status field in wharf-web's project view (Failed/Completed)
+currently calls an endpoint that asks the server to fetch all TRX files
+for the build, and take the most recent one. Then it parses the file to
+see how many failed/completed tests there were.
+It does this for every build in the view, every time it is refreshed.
+This leads to unnecessary requests and computation. This, in turn, slows
+down the response times and increases the API load.
```
applejag, line 48: (open question) Would you regard tackling the "n+1" issue related to this RFC or out-of-scope to this RFC?

When fetching the list of builds, we don't want to also fetch the test result summaries on a request per build, but instead get that info in the build list directly. Such as having the summed up summaries in the build object returned by the `GET /project/{projectid}/builds` endpoint

Alexamakans: Feels like it is within scope of the original idea of optimizing the process/flow. Although the Summary I've written is more focused on the filtering/parsing, that's more because I didn't consider it before.

_published/0014-test-results-handling.md
applejag: Don't forget to add jemoji: https://docs.github.com/en/enterprise/2.13/user/articles/emoji-on-github-pages

Alexamakans: Added just now in https://github.com/iver-wharf/rfcs/pull/14/commits/ed272c479dc4024440b2e59b98121f8ea3012251

_published/0014-test-results-handling.md
applejag: Showing a glimpse of the test summary in the builds list is one of the big requirements for this feature, so I think it would be good then to elaborate on that in the RFC

_published/0014-test-results-handling.md
```diff
@@ -241,7 +242,7 @@ Nothing comes to mind.
 Nothing comes to mind.
 
 ## Unresolved questions
-
+might
```
applejag:
```suggestion

```

applejag (approved):

_published/0014-test-results-handling.md
```diff
@@ -0,0 +1,280 @@
...
+## Alternative solutions
+
+Nothing comes to mind.
+
+## Future possibilities
+
+Nothing comes to mind.
+
+## Unresolved questions
+Nothing comes to mind.
```
Alexamakans, line 246:
```suggestion

Nothing comes to mind.
```

fredx30 (approved):
Lgtm at this point but i have been less involved in the discussions up to here as im just joining in now.

Alexamakans:
@Pikabanga Can't merge with the requested changes.
Would you mind taking a look to see if it's been resolved?

Not sure if you've received a notification about the recent commits, so I'm notifying through this message.

Pikabanga (approved):

Pikabanga:
> @Pikabanga Can't merge with the requested changes.
> Would you mind taking a look to see if it's been resolved?
> 
> Not sure if you've received a notification about the recent commits, so I'm notifying through this message.

Approved!

PR #17, opened 2021-08-25
AzureDevOps repos, and not projects

applejag:
Resolving two birds with one RFC here:

1. We're importing each Azure DevOps project as its own Wharf project, while a
   project in Azure DevOps may contain more than one Git repositories. To Wharf,
   these repositories should be treated as distinct Wharf projects.

2. When importing Azure DevOps projects from <https://dev.azure.com/>, the Git
   SSH URL is assigned an invalid value, as reported over in
   <https://github.com/iver-wharf/wharf-provider-azuredevops/issues/24>.

   This because we're importing projects and not the repositories so we don't
   get a Git SSH URL from the remote API and instead try to construct it on an
   inaccurate assumption, as can be seen here:
   [github.com/iver-wharf/wharf-provider-azuredevops/blob/internal/importer/importer.go](https://github.com/iver-wharf/wharf-provider-azuredevops/blob/7b6397029b9bbe10e14e1367195e2491bd6eae83/internal/importer/importer.go#L274-L285)

## Rendered

https://github.com/iver-wharf/rfcs/blob/rfc/azuredevops-repos-and-not-projects/_published/0017-azuredevops-repos-and-not-projects.md


_published/0017-azuredevops-repos-and-not-projects.md
```diff
@@ -0,0 +1,259 @@
...
+   GET https://dev.azure.com/{organization}/_apis/git/repositories/{repositoryId}/refs?filter=heads/
+   ```
+
+   <https://docs.microsoft.com/en-us/rest/api/azure/devops/git/refs/list#refs-heads>
+
+   This is used to get the following information:
+
+   - Branch name (but make sure to trim away the `refs/heads/` prefix)
+
+## Compatibility
```
fredx30, line 163: This deserves a note. My gut feeling was that there would be two different types of importers for local instances or private cloud and public cloud instances of azure dev ops. 

Does this functionality currently work on private cloud instances? Will it continue working on private cloud instances after this change? Does this RFC convert a private cloud provider into a public cloud provider because the two are incompatible?

applejag: Excellent question! I'll add a well deserved note :)

_published/0017-azuredevops-repos-and-not-projects.md
```diff
@@ -0,0 +1,259 @@
...
+  The latter, naming C, is used in wharf-provider-azuredevops v1. For v2, how
+  do we handle the naming when there are more than 1 code repository?
+
+  1. Always use the naming A.
+
+  2. Use the naming C, unless the Azure DevOps project contains more than 1
+     code repository, where naming A should be used instead.
+
+  3. Use the naming B/C when the `{azure-repo}` = `{azure-project}`, and use
+     naming A otherwise.
```
fredx30: Could you null check the string and concatenate them such that the project group collumn becomes the name? Id dosent seem like a human friendly name.

_published/0017-azuredevops-repos-and-not-projects.md
```diff
@@ -0,0 +1,259 @@
...
+possible declaring these impacts. We should also write a document on
+<https://iver-wharf.github.io/> about upgrading wharf-provider-azuredevops from
+v1 to v2, where we explain these backward incompatibilities as well.
+
+## Alternative solutions
+
+For the sake of resolving
+<https://github.com/iver-wharf/wharf-provider-azuredevops/issues/24> we could
+settle with still importing 1 Wharf project per Azure DevOps project, but
+this is so tightly coupled so I [@jilleJr] think it's best to tackle this
```
fredx30: nit: Form here as you are the author of the RFC and this is a techincal doc opinions should be reformulated as fact. 
I think debates should go into the `Unresolved questions`, the RFC should assume this debate has been done. The alternatives considered has always seem to be more of a inventory check to me. As in if we come back to this later and ask why didnt we consider x? Then we get that answer here in the form of a benefits drawback discussion or alternativly we get the answer "we didnt consider x at all".

_published/0017-azuredevops-repos-and-not-projects.md
applejag: Sorry for the confusion. The `{var}` fields here represent the different names, and not the IDs/GUIDs. I will add the following snippet that I added to the "Wharf project naming" section:

---

Where:

- `{azure-org}` = name of the Azure DevOps organization that holds the Azure
  DevOps project.

- `{azure-project}` = name of the Azure DevOps project.

- `{azure-repo}` = name of the code repository inside the Azure DevOps project.

_published/0017-azuredevops-repos-and-not-projects.md
applejag: I like this proposed approach. Personally I've considered the "Alternatives" section as pointing out the obvious alternatives (to begin with) so they can be dismissed during the review period. Were the author answers potential "why didn't you just do X?" questions.

But if I understand you correctly, we could also move the "unresolved questions" over to "alternatives" once they're answered/discussed? Because I like that.

_published/0017-azuredevops-repos-and-not-projects.md
applejag: Note added. At the time of writing, this is what I came up with:

### Keeping compatibility with self-hosted instances

Will be kept. Compared to v1, we will use 1 new endpoint:

```http
GET https://dev.azure.com/{organization}/{projectId}/_apis/git/repositories
```

This new endpoint does also exist in Azure DevOps API v5.0, which we have
relied on up until now.

As for the new procedure of getting the Git SSH URL, this will produce better
stability than before for unconventional installations of self-hosted
Azure DevOps instances as long as their Git SSH URL is configured correctly,
whereas conventional installations will continue to behave as before.

_published/0017-azuredevops-repos-and-not-projects.md
applejag: I don't fully understand what you mean with

> Could you null check the string and concatenate them such that the project group collumn becomes the name?

None of the names of the Azure DevOps org, project, nor repo can be null. The idea was to concatenate the org and the project names (with a slash `/` inbetween) in some cases, to be used as the Wharf group name. Similar to how we have concatenated subgroups in the wharf-provider-gitlab where a project at `https://our-internal-gitlab-instance.example.com/topgroup/subgroup/projectname` would result in the Wharf group name `topgroup/subgroup` and the Wharf project name would resolve to `projectname`.

fredx30: To me it appears the project-group collumn is a always available and unique name whereas the project name collumn does not appear unique. I think the uniqueness is important when viewing from wharf, lack of uniqueness seems like it could cause confusion during project selection.

Why use a slash to concatenate the names if the name is to be part of a URL shouldnt `/` then be designated an illegal character?

_published/0017-azuredevops-repos-and-not-projects.md
fredx30: I think it would depend on the nature of the unresolved question but in principal yes. If the unresolved question is expected to be classified as an alternative solution in advance- i think i would expect a discusssion call/brainstorming meeting. Listing it as a discussion for a text forum such as a PR seems inefficient. Meetings are great for brainstorming. Once the upsides and downsides are listed as facts it could then go straight into the alternatives considered, or in the  alternative case replace the old core idea as the target of the RFC.

_published/0017-azuredevops-repos-and-not-projects.md
```diff
@@ -0,0 +1,259 @@
...
+  | ID  | Project group                 | Project name     |
+  | --- | -------------                 | ------------     |
+  | A   | `{azure-org}/{azure-project}` |`{azure-repo}`    |
+  | B   | `{azure-org}`                 |`{azure-repo}`    |
+  | C   | `{azure-org}`                 |`{azure-project}` |
+
+  The latter, naming C, is used in wharf-provider-azuredevops v1. For v2, how
+  do we handle the naming when there are more than 1 code repository?
+
+  1. Always use the naming A.
```
iverestefans: I am voting for first option. Adding conditions to naming as it is proposed in 2nd and 3rd points can make it less readable and generate more bugs.

iverestefans (approved):
LGTM!

_published/0017-azuredevops-repos-and-not-projects.md
applejag: You're right. My only intention with this "conditional naming" was so the previously imported Wharf projects keep their names, and to try and avoid double naming. But if it's consistently double-naming then at least it's easier to understand for end-users.

I'm fond of this argument, so I'll update the RFC to use option 1.

_published/0017-azuredevops-repos-and-not-projects.md
applejag: I'll call you so we can resolve this verbally. I think I might need to rubber-duck this a little as well :)

_published/0017-azuredevops-repos-and-not-projects.md
applejag: Yea cool, so I think we have the same mindset here. Correct me if I misunderstood, but here's my image:

- "Alternatives": document some resolved questions and other obvious misdirections the author intentionally didn't take
- "Unresolved questions": things to talk about in meeting or discuss in PRs, preferably verbally for speed and brainstorming

_published/0017-azuredevops-repos-and-not-projects.md
applejag: Updated

_published/0017-azuredevops-repos-and-not-projects.md
fredx30: I think the relevance of alternatives increase in the scenario where there are multiple good mutually exclusive options to move forward with. The discussion should be included for the sake of reference.

_published/0017-azuredevops-repos-and-not-projects.md
applejag: For ref: We cleared up the confusion

- For all our currently supported remote providers, the group can never be null. Explained in detail over at the `REPO_GROUP` env var documentation: https://iver-wharf.github.io/#/usage-wharfyml/variables/built-in-variables?id=repo_group

  > - For GitLab: Namespace (user, group, or group/subgroups)
  > - For AzureDevOps: Organization or collection name.
  > - For GitHub: Organization name or username.

  After this RFC, the Azure DevOps will instead look like this:

  > - For AzureDevOps: Organization name and project name, separated by slash `/`.

- A future enhancement might be to normalize the Wharf group names so we can have them in the URL without custom/convoluted routing logic. Perhaps by replacing `/` with `-`. Leaving that for a different day though.

- Another future plan, which is referenced in https://github.com/iver-wharf/wharf-provider-azuredevops/issues/6 and https://iver-wharf.github.io/wharf-notes/hide-providers-behind-api is to start using the remote IDs for mapping projects instead of this "by-name basis" that we have now. But that work is already ongoing

Alexamakans (approved):
👍

applejag:
Created issue: https://github.com/iver-wharf/wharf-provider-azuredevops/issues/30

PR #20, opened 2021-09-14
Added issue to RFC-0014

Alexamakans:
Had forgotten to create and add the issue for RFC-0014, this fixes that.﻿


applejag (approved):

fredx30 (approved):
Kudos to the cleanup :)

PR #21, opened 2021-09-15
Change from issue to milestone

applejag:
This is an experiment, but what about using a milestone when implementation is easier to handle as divided issues instead of one big issue that references all child issues?

I'm not :100: on this, but I think it's a nice thing to have. https://github.com/iver-wharf/wharf-api/milestone/1

Note: I'm not proposing that we all do this. But if it's nice, then maybe you others will adopt it as well. Fully optional. If it works well, then maybe we could incorporate the functionality into the 0000-template.md as well.

Rendered preview:

Screenshot from 2021-09-15 16-23-28


Alexamakans (approved):
This seems like a nice way to organize it.
👍 from me.

Pikabanga (approved):

PR #22, opened 2021-09-15
Added header styling

applejag:
I found that in some pages when reading RFCs it was difficult to distinguish between the different header levels, as well as distinguishing between header and regular paragraph in some cases.

This adds a bottom border and some `###` before the headers to make them stand out more.

Also sneaked in a `color-scheme` setting to tell the web browser that it's already dark theme'd, as my browser wants to make it darker in some senses, such as inverting the white bottom border to a black border, which didn't fit at all.

## Preview

Before:

Screenshot from 2021-09-15 18-47-33

After:

Screenshot from 2021-09-15 18-47-20

This page is used in the above screenshots: https://iver-wharf.github.io/rfcs/published/0017-azuredevops-repos-and-not-projects

### DIFF
--- a/_sass/custom/custom.scss
+++ b/_sass/custom/custom.scss
@@ -1,39 +1,39 @@
 :root {
-	color-scheme: dark;
+  color-scheme: dark;
 }
 
 h1::before {
-	color: #f5f6fa;
-	content: '# ';
-	font-family: "SFMono-Regular", Menlo, Consolas, Monospace;
+  color: #f5f6fa;
+  content: '# ';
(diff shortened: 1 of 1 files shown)


Alexamakans (approved):
nit: I would prefer if `h1` had no prefix, starting with one character at `h2` and so on.

nit2: I find `#` pretty big and distracting, maybe something like `-` or `>` would be better?
`#` is good in raw format because it's really in your face and makes it obvious, but when reading the rendered format it feels more in the way because you have other things that tell you what level you're at, like font size and weight of the header.

iverestefans (approved):
LGTM!

PR #25, opened 2021-11-19
RFC: wharf-cmd provisioning

applejag:
Our previous solution was oriented around Jenkins. This has been proven unstable and difficult for various reasons, such as deployment of Wharf and Jenkins together in Kubernetes, debugging capabilities, ability to execute `.wharf-ci.yml` builds locally, and relaying build logs and artifacts to the wharf-api and ultimately end-user, etc. This RFC describes our replacement of Jenkins with four new Wharf components.

There is a heavy focus on making the wharf-cmd-worker subcommand perform Wharf builds standalone, without a wharf-api and wharf-web to present the results.

## Rendered

https://github.com/iver-wharf/rfcs/blob/rfc/wharf-cmd-provisioning/_published/0025-wharf-cmd-provisioning.md


fredx30 (approved):
Thanks for good discussions around this! =)
Good work. Will checkout again some other day if I remember or think of something else.

Alexamakans (approved):

PR #26, opened 2021-12-13
RFC: v2+ Go modules release

applejag:
Go modules, which is the package management strategy we use in Wharf, has
strict rules on how to publish v2+ releases. We have not been Go-modules
compliant in some of our repositories. This RFC suggest one of the many
alternative solutions for being compliant, with its up- and downsides.
This is an important decision, as it induces a lot of headaches if we do it
wrong and have to switch to a different strategy later on.

This is based on a conversation that popped up in a PR comment thread:
<https://github.com/iver-wharf/wharf-api-client-go/pull/29#discussion_r766513730>

## Rendered

https://github.com/iver-wharf/rfcs/blob/rfc/go-modules/_published/0026-v2-go-modules-release.md

### DIFF
--- a/_published/0026-v2-go-modules-release.md
+++ b/_published/0026-v2-go-modules-release.md
@@ -87,7 +87,7 @@ Contrary to some of the official Go recommendations, we do not apply any of the
 following strategies:
 
 - :warning: No duplicating code into subdirectories, where v1 would be kept at
-  `<repo root>/` and v2 would be kept at `<repo root>/v2/`. This breaks
+  `<repo_root>/` and v2 would be kept at `<repo_root>/v2/`. This breaks
   compatibility with Go versions older than 1.9.7, 1.10.3, and 1.11.0.
 
 - :warning: No separate branches for different major versions. All code will
@@ -122,7 +122,7 @@ The above decisions has been made from researching the following pages:
 ## Compatibility
 
 This breaks compatibility with developers using vendoring (where copies of the
-dependency source code is stored in `<repo root>/vendor/...`) and/or GOPATH
+dependency source code is stored in `<repo_root>/vendor/...`) and/or GOPATH
 (where source code you're working on is kept in `$GOPATH/src/...`, eg.
 `~/src/...`).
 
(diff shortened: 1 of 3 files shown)


Alexamakans (approved):

_published/0026-v2-go-modules-release.md
```diff
@@ -0,0 +1,180 @@
...
+> longer than the full [Go language specification](https://go.dev/ref/spec),
+> but worse is that the Go modules resources are spread out across blog posts,
+> wiki's, and documentation sites.
+
+This is based on a conversation that popped up in a PR comment thread:
+<https://github.com/iver-wharf/wharf-api-client-go/pull/29#discussion_r766513730>
+
+## Motivation
+
+The wharf-api is an example of a version 2 or higher Go module we've release.
```
Pikabanga:
```suggestion
The wharf-api is an example of a version 2 or higher Go module we've released.
```

_published/0026-v2-go-modules-release.md
```diff
@@ -0,0 +1,180 @@
...
+
+ import (
+-    "github.com/iver-wharf/wharf-api/pkg/model/database"
++    "github.com/iver-wharf/wharf-api/v5/pkg/model/database"
+ )
+```
+
+### Counter-recommendation decisions
+
+Contrary to some of the official Go recommendations, we do not apply any of the
```
Pikabanga: This phrasing confuses me, so we do not apply **no** separate branches?

We don't use separate branches for major versions right? The double negation makes it sound like we do use separate branches. But maybe that's just me.

Pikabanga (changes requested):

_published/0026-v2-go-modules-release.md
applejag: No you're completely right. This paragraph is a mess with its double negation.

I've changed it:

From:

> Contrary to some of the official Go recommendations, we do not apply any of the following strategies:

To:

> Contrary to some of the official Go recommendations, our practices are:

Pikabanga (approved):

PR #27, opened 2021-12-14
Added styling for blockquotes

applejag:
The lack of styling for block quotes was annoying me. It was very unclear that it was a blockquote. This adds much more emphasis on the fact that there's a block quote there.

The weird mix of using `em` and `px` is because I wanted to use the same alignment and border radius settings as the other components such as the tables and code blocks.

The left border is there to distinguish from nested blockquotes.

### Before

Screenshot from 2021-12-14 13-14-34

### After

Screenshot from 2021-12-14 13-14-16


Alexamakans (approved):
LGTM 👍

fredx30 (approved):

PR #28, opened 2022-01-11
RFC: Repo labels

applejag:
We are currently using mostly "vanilla GitHub" labels. This restricts us in our
way to quickly convey the categories of issues and PRs, as well as our lack of
prioritization makes it really tough to understand what is important.

## Rendered

https://github.com/iver-wharf/rfcs/blob/rfc/repo-labels/_published/0028-repo-labels.md


applejag:
Closing after an internal discussion where we concluded that I (@jilleJr) is the only one needing this for my project planning and work delegation, so I do not need an RFC for it.

Will not implement any of the other label categories (type, size) as they are not deemed needed yet.

